PR #4, opened 2019-12-30
Feature/endpoint paginator

davecoates:
Pagination usage (will go into wiki once finalised) (in the testbed this is wrapped up in `useEndpoint`)

```js
// Where the state for pagination will be synced to
const [paginationState, setPaginationState] = useState();
// Create the paginator once. We pass the `setPaginationState` function which is called whenever
// pagination state changes. This triggers a re-render and API will be called.
const paginator = useMemo(() => endpoint.createPaginator(null, setPaginationState), [endpoint]);

// Make sure state is always in sync. This is particularly useful when the state can change external
// to paginator, eg. when state comes from URL
paginator.syncState(paginationState);

// Prepare takes into account the state of the paginator - when paginator state changes it returns
// a new action.
const preparedAction = endpoint.prepare({ paginator });

// Fire the endpoint initially and whenever it changes.
const { data } = useSWR([preparedAction], action => action.execute())
```

Still to do:

1) If an API call fails then pagination state is essentially invalid. This isn't straightforward due to how divorced the state handling and execution of the action is.

2) Similarly if multiple changes to pagination occur before first resolves responses could come back out of order and pagination state will be out of sync with results we have

3) Documentation on creating paginators / customising how responses are handled for different API implementations

---

I was also a bit worried about having state in the paginator itself and then syncing it externally but have no better solution right now. I was particularly worried about how it might behave in concurrent mode but I think it's fine if you call `syncState` 🤷‍♂ 

Open to better solutions here but at this stage ready to proceed with this and then revisit it once we start using it properly.


django-testbed/django-root/xenopus_frog/views.py
```diff
@@ -16,7 +19,29 @@ class Meta:
...
+
+
 class UserViewSet(ModelViewSet):
     queryset = User.objects.all().order_by("first_name", "last_name", "email")
     serializer_class = UserSerializer
     permission_classes = []
+
+    @property
+    def pagination_class(self):
+        pagination_type = self.request.query_params.get("paginationType")
```
levic, line 57: Is the ability for the client to choose pagination just for demonstration purposes?

If so, probably should add a comment to that effect: it's not common to allow the client to choose this and would be a backwards compatibility burden if you were exposing a public API

levic:
Extra comments for example usage:
- `useState()` has no explicit defaults because the default implementation changes the pagination method according to what the server returns

levic:
- Am unclear where the module boundaries are around `EndPoint`, `useEndPoint` and `usePaginator` are: what goes in your app and what comes from the library?

levic:
- Is `syncState` necessary? It can't be part of the pagination creation automatically because of the memo? (Or does it not really matter because it's expected that the app will contain a `useEndpoint` or equivalent?)

django-testbed/frontend/src/views/UserListView.js
```diff
@@ -1,30 +1,66 @@
...
+import qs from 'qs';
 
 import User from '../models/User';
 import useConnected from '../useConnected';
 import useEndpoint from '../useEndpoint';
 import UserCreateUpdateView from './UserCreateUpdateView';
 
 export default function UserListView() {
+    const { search, pathname, origin } = window.location;
+    const { paginationType } = qs.parse(search.split('?').pop());
```
levic: Minor style issue: I find the latter to be clearer (I don't see why qs doesn't make it the default though; if you're embedding `?` in your query strings without url encoding you're doing it wrong)
```js
const { paginationType } = qs.parse(search.split('?').pop());
const { paginationType } = qs.parse(search, { ignoreQueryPrefix: true });
```

levic: As above: I assume the user-selected pagination type is just for demo purposes?

js-packages/@prestojs/rest/src/CursorPaginator.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    constructor(
+        initialState: CursorPaginationState = {},
+        onChange?: PaginationStateChangeListener
+    ) {
+        super(initialState || {}, onChange);
+    }
+
+    setPageSize(pageSize: number): void {
```
levic: Do you need to allow for `null`/`''` (meaning default)?

js-packages/@prestojs/rest/src/CursorPaginator.ts
```diff
@@ -0,0 +1,116 @@
...
+        if (pageSize < 1) {
+            throw new Error(`Invalid pageSize ${pageSize} - should be >= 1`);
+        }
+        this.setState(currentState => {
+            // If page size changes any next/previous cursors are no invalid
```
levic: Is that actually the case? If you're using a cursor then the page size shouldn't matter since all the cursor-based schemes I've encountered only use the cursor as a "bookmark" and you can fetch as many prior/after that index as you want. You might get some quirkiness but it should still work

(eg with page size 20 if you do Next -> Next -> Next, then change page size to 100: you're now seeing items 60-159; now choose Prev and you're seeing 0-99 which overlaps with 60-159, but it still works)

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -270,6 +290,11 @@ export default class Endpoint {
         this.requestInit = requestInit;
     }
 
+    createPaginator(initialState = {}, onChange?): PaginatorInterface {
```
levic: What's this function used by? I only see it used by test cases

Edit: ignore, was hidden by github collapsing things

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -326,32 +355,71 @@ export default class Endpoint {
...
+    async execute(options: EndpointExecuteOptions = {}): Promise<ExecuteReturnVal> {
+        if (options.paginator) {
+            options = options.paginator.getRequestInit(options);
+        }
+        const { urlArgs = {}, query, paginator, ...init } = options;
         const url = this.urlPattern.resolve(urlArgs, { query });
         try {
             const cls = Object.getPrototypeOf(this).constructor;
-            return fetch(
+            const requestInit = mergeRequestInit(
```
levic, line 406: Doesn't really belong to this line, but the implementation of `mergeRequestInit` has a [dangling comment](https://github.com/prestojs/prestojs/pull/4/files#diff-e167769c3664d0e03f93b134824c4419R44) before the header

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -326,32 +355,71 @@ export default class Endpoint {
...
                     return res;
                 })
                 .then(res => this.decodeBody(res))
-                .then(data =>
-                    this.transformResponseBody ? this.transformResponseBody(data) : data
-                );
+                .then(data => {
+                    returnVal.decodedBody = data;
+                    if (paginator) {
+                        // If we have a paginator update it's state based on response
```
levic: `it's` => `its`

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -326,32 +355,71 @@ export default class Endpoint {
...
+                            paginator,
+                            returnVal
+                        );
+                        if (paginationState) {
+                            paginator.setResponse(paginationState);
+                            data = paginationState.results;
+                        } else {
+                            // TODO: If you specify a paginator and the response is not paginated should
+                            // we warn? I think yes as I can't think of a reason why not but may need to
+                            // revisit this if we find a usecase
```
levic, line 443: What if there are no pages? eg you specify cursor-based pagination but there is only 1 row then you might not have a Next/Prev link available (or are you assuming they'd be empty?)

js-packages/@prestojs/rest/src/CursorPaginator.ts
```diff
@@ -0,0 +1,116 @@
...
+        this.setState(currentState => {
+            const nextState: CursorPaginationState = { ...currentState };
+            delete nextState.cursor;
+            return nextState;
+        });
+    }
+
+    next(): void {
+        if (this.nextCursor == null) {
+            throw new Error('Cannot go to next page as next cursor is not known');
```
levic: Is there any way to structure this so it can return a link URL instead of exposing a function that performs the page navigation?

My concern is that you have no way of creating a regular `<a href="...">` link because you don't know what URL you're going to be navigating to (although the fact that the way state is stored is opaque to the paginator may preclude this?)

(This also applies to all of the other places where pagination navigation occurs)

js-packages/@prestojs/rest/src/InferredPaginator.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    get currentState(): PaginationState | null | undefined {
+        return this.paginator?.currentState;
+    }
+
+    get total(): number | null | undefined {
+        if (this.paginator) {
+            if (this.paginator instanceof CursorPaginator) {
+                throw new Error('total not valid for cursor pagination');
+            }
```
levic: Rather than throwing an error, is it going to be easier for clients to just get a `null` back?

Am thinking about a listing screen where you don't know the pagination. It would be more convenient to do something like
`{paginator.total != null && <span>Total: {paginator.total} records</span>}` than to have to catch an exception. It also means that the same logic can handle results that simply haven't come back from the server yet

levic: (This applies to most of the InferredPaginator methods)

js-packages/@prestojs/rest/src/LimitOffsetPaginator.ts
```diff
@@ -0,0 +1,131 @@
...
+    ) {
+        super(initialState || {}, onChange);
+    }
+
+    setLimit(limit: number): void {
+        if (limit < 1) {
+            throw new Error(`Invalid limit ${limit} - should be >= 1`);
+        }
+        this.setState(currentState => {
+            // When page size changes we need to reset the offset
```
levic: As with previous comment in CursorPaginator, don't need to do a full reset.

If you want to remove any quirks then can just round offset to the nearest integer multiple of limit (this is what bazooka does)

levic: Specifically, as a user if I'm on page 18 and decide "this is too slow, I want to see 100 at a time" then I don't want to be sent back to the start again

js-packages/@prestojs/rest/src/LimitOffsetPaginator.ts
```diff
@@ -0,0 +1,131 @@
...
+                const nextState: LimitOffsetPaginationState = { ...currentState, limit };
+                delete nextState.offset;
+                return nextState;
+            }
+            return { ...currentState, limit };
+        });
+    }
+
+    setOffset(offset: number): void {
+        if (offset < 0) {
```
levic: As before; should allow `null` or `''` to reset to default?

davecoates:
> * Am unclear where the module boundaries are around `EndPoint`, `useEndPoint` and `usePaginator` are: what goes in your app and what comes from the library?

Currently `Endpoint` is the only thing in the library. There's no such thing as `usePaginator`, at least not yet.

I think eventually we'll add `useEndpoint` to the library but only after using it a bit in the testbed to workout what it should look like

js-packages/@prestojs/rest/src/LimitOffsetPaginator.ts
```diff
@@ -0,0 +1,131 @@
...
+            throw new Error(`Invalid offset ${offset} - should be >= 0`);
+        }
+        if (this.total != null && offset >= this.total) {
+            throw new Error(
+                `Invalid offset ${offset} - should be less than the total number of results (${this.total})`
+            );
```
levic: Total could be out of date: can we not let this be something that the server is responsible for dealing with? If I ask for offset 9999 but there's only 120 records then the server can say "here are your results, but I used offset 100" (or alternately, "here are 0 results")

levic: (Unnecessary restrictions also apply to next() & prev())

davecoates:
> * Is `syncState` necessary? It can't be part of the pagination creation automatically because of the memo? (Or does it not really matter because it's expected that the app will contain a `useEndpoint` or equivalent?)

I'd like it not to be but don't see a way around it. The issue is that the paginator class itself is all handles outside of React. It makes state changes then gets sync'd to React by the onChange callbacks. This is only one way though - to get changes from React (eg. if the URL changed when reading state from there) need some way to communicate that back. 

It's the part of this I'm most concerned about and like the least... but not sure what to do about it yet. Hopefully through usage something better will become clear...

js-packages/@prestojs/rest/src/getPaginationState.ts
```diff
@@ -0,0 +1,79 @@
+import qs from 'qs';
+import { ExecuteReturnVal } from './Endpoint';
+import InferredPaginator from './InferredPaginator';
+import Paginator from './Paginator';
+
+/**
+ * Return the state for a paginator based on the response itself.
+ *
+ * The default implementation is for is in conjunction with InferredPaginator where the pagination type
```
levic: `is for is in` => `is for use in` ?

js-packages/@prestojs/rest/src/getPaginationState.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (Array.isArray(decodedBody.results) && typeof decodedBody.count == 'number') {
+        let { pageSize } = decodedBody;
+        if (!pageSize) {
+            const { page } = query || {};
+            // Infer page size if we are on first page and total results are greater
+            // than returned results
+            if (decodedBody.count > decodedBody.results.length && (page === 1 || page == null)) {
+                pageSize = decodedBody.results.length;
+            }
+        }
```
levic, line 69: If I have a page size of 20 and 23 total records, and immediately navigate to page 2 without going to page 1 first then it won't know the page size?

js-packages/@prestojs/rest/src/getPaginationState.ts
```diff
@@ -0,0 +1,79 @@
...
+    const isLimitOffsetLink = (link: any): boolean =>
+        typeof link == 'string' && link.includes('limit=');
+    if (isLimitOffsetLink(decodedBody.next) || isLimitOffsetLink(decodedBody.previous)) {
+        const { limit } = qs.parse((decodedBody.previous || decodedBody.next).split('?').pop());
+        return {
+            total: decodedBody.count,
+            results: decodedBody.results,
+            limit: Number(limit),
+        };
+    }
```
levic, line 56: Am thinking about how most REST endpoints don't include a prev/next link, they just assume you passed in an offset/limit (or page number) and then return the records you asked for; any pagination control is essentially done client-side

Is it worth also checking the request to look for a limit/offset, or at least for DRF support we don't care and can implement that later?

(Edit: this is assuming the client doesn't just specify a specific pagination method instead of using inference)

levic:
Potential high-level changes:
- Throwing errors in InferredPaginator vs returning null
- Doesn't seem to be possible to create regular <a> links to do paginated navigation
- Shouldn't do a full reset of paginator when changing page size
- I don't think we need to be so strict with limit/offset/pagesize restrictions (if data isn't back yet, or totals are out of range then we can still navigate to a page that may have empty results)
- Inference of pagination: do we need to add support for endpoints that don't return any pagination-related info at all? (ie we look at the request)

davecoates:
> Potential high-level changes:
> 
> * Throwing errors in InferredPaginator vs returning null

This essentially would ignore incorrect usage though wouldn't it? What's the motivation? 

> * Doesn't seem to be possible to create regular  links to do paginated navigation

Not sure I follow why this wouldn't work - you'd just initialise the paginator state from the URL on load.

> * Shouldn't do a full reset of paginator when changing page size

Why? If page size changes then the page you were on has a different meaning and could be invalid which would result in an error (even if we remove the restriction on accessing invalid pages the backend will still 404 in the case of DRF)

> * I don't think we need to be so strict with limit/offset/pagesize restrictions (if data isn't back yet, or totals are out of range then we can still navigate to a page that may have empty results)

You specifically can't do some things if the totals aren't back (eg. impossible to go to last page). Do you mean to just silently ignore the call entirely in those cases and in the other cases just allow it?

I feel silently ignoring it is bad - probably indicates your UI is not user friendly (it should be disabled until the user can interact with it). I don't have a problem allowing invalid states as that should result in feedback to the user one way or the other.

> * Inference of pagination: do we need to add support for endpoints that don't return any pagination-related info at all? (ie we look at the request)

We already do that - see `getPaginationState.ts`

levic (commented):
> > * Shouldn't do a full reset of paginator when changing page size
> 
> Why? If page size changes then the page you were on has a different meaning and could be invalid which would result in an error (even if we remove the restriction on accessing invalid pages the backend will still 404 in the case of DRF)

Can you give me an example where it would fail?

- CursorPaginator: _may_ fail if is cursor state is stored server-side and maintained between requests. If the cursor is actually just an index to fetch the next set of records from then continues to work fine
- LimitOffsetPaginator: works fine if you change the page size. May want to round offset to a multiple of limit if changing page size
- PageNumberPaginator: need to adjust current page number so that you are in the same spot but other than that works fine to change the page size

levic:
> > Potential high-level changes:
> > 
> > * Throwing errors in InferredPaginator vs returning null
> 
> This essentially would ignore incorrect usage though wouldn't it? What's the motivation?
> 

There's a few use cases:
- You're trying to call the wrong kind of paginator method on an InferredPaginator. In this case it is indeed incorrect usage, but the motivation to ignore is that it's a little more convenient to deal with null than an exception in JSX. Mostly it means that UI code that deals with all 3 pagination types (eg a generic List View) is simpler: it just hides links that don't work. (Given the infrequency at which this code is likely to happen this may be a weak justification)
- You're calling the right kind of paginator method on an InferredPaginator, but data hasn't been returned from the server yet. In this case, again it's easier to deal with a null return value where you just eg hide the "Last" link than to deal with an exception.
- You're calling the right kind of method but with the wrong parameter: eg a page number should never be <1. A total may be off the end though: if the last record is deleted and you click "Next" and now go to a page with no results then should the user see an error? I'd expect no, they'd just see no results.

> > I don't think we need to be so strict with limit/offset/pagesize restrictions (if data isn't back yet, or totals are out of range then we can still navigate to a page that may have empty results)
You specifically can't do some things if the totals aren't back (eg. impossible to go to last page). Do you mean to just silently ignore the call entirely in those cases and in the other cases just allow it?
>
> I feel silently ignoring it is bad - probably indicates your UI is not user friendly (it should be disabled until the user can interact with it). I don't have a problem allowing invalid states as that should result in feedback to the user one way or the other.

"it should be disabled until the user can interact with it" -- how can we know this with an InferredPaginator though? We don't know whether a "Last" link is even going to be possible until we get results back from the server, so we're going to have to hide the link rather than showing a loading spinner/disabled on it. If we're doing that then we should return null to make it simpler for the consumer of the API.

Perhaps the answer is for some methods of InferredPaginator to never throw an exception and always return null whereas the underlying Paginator throws an exception on data that's never going to be valid. Data that simply isn't ready yet (as opposed to 'will _never_ be ready') would also return null

levic:
> > * Doesn't seem to be possible to create regular  links to do paginated navigation
> 
> Not sure I follow why this wouldn't work - you'd just initialise the paginator state from the URL on load.
> 

I'm talking about generating a DOM that looks something like:

```
    <a href="http://mysite.com/users/?page=2&size=20">Next</a>
    <a href="http://mysite.com/users/?page=4&size=20">Prev</a>
    Results
    <a href="http://mysite.com/users/?page=6&size=10">10</a>
    <a href="http://mysite.com/users/?page=3&size=20" class="selected">20</a>
    <a href="http://mysite.com/users/?page=2&size=40">40</a>
    <a href="http://mysite.com/users/?page=1&size=100">100</a>
```

How can I get the URLs for these links? The API exposes methods to _go to_ those URLs, but it does this via `setState()` which in turn sets the URL state. There's no way for a consumer of the API to know the URL without navigating there first

django-testbed/django-root/xenopus_frog/views.py
davecoates: Just demo. Anything in `django-testbed` is test code.

django-testbed/frontend/src/views/UserListView.js
davecoates: Ah wasn't aware of that option.

Yes - demo (it's in `django-testbed`)

davecoates: Probably test code is better than demo - it's not really for showing anyone apart from devs working on xenopus to test features outside of test cases

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: I'm pretty sure I just meant in terms of consistent page sizes it's no longer valid (ie. it now points to something part way down a page, so if you go previous back to the start you'll get an overlap on page 1 and page 2)

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: Can you expand on this more because I don't really follow? 

Are you talking about returning a link to the _api_ or to a link to the frontend that then calls the api? I see no reason you can't do the latter (there's an example of how you do this although it relies on a hook that you'd currently have to write - search for `useUrlState`)

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: ta - looks like bad merge

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: it would be empty in that case

this is for where the response itself is definitely not a paginated shape

I _think_ we will probably want to remove this but I'll leave it for now until the case comes up. I suspect we'll want to write a hook or generic UI component that won't know upfront and so will create a paginator in case it's required

js-packages/@prestojs/rest/src/InferredPaginator.ts
davecoates: I was thinking that in most cases you know what pagination you are dealing with and so you'd write your UI for one type of pagination so I was erring on the side of letting user know about misuse rather than silently ignoring it.

For framework/generic code I would have thought you'd write the pagination UI as a component and switch it based on the inferred type, not based on the presence of each variable (eg. you'd have next/previous for cursor based but for a lot of page number based interfaces if they have next/previous at all they are to either side of pages)

eg.

```jsx
let paginationEl = null;

// If we know pagination type...
if (paginator.paginator) {
   if (paginator.paginator instanceof LimitOffsetPaginator) {
      paginationEl = <LimitOffsetPaginatorButtons paginator={paginator} />;
  }
   if (paginator.paginator instanceof PageNumberPaginator) {
      paginationEl = <PageNumberPaginationButtons paginator={paginator} />;
  }
   if (paginator.paginator instanceof CursorPaginator) {
      paginationEl = <CursorPaginatorButtons paginator={paginator} />;
  }
}
```

So I am tentatively on the side of current approach but can be convinced otherwise if you feel it's too strict

js-packages/@prestojs/rest/src/getPaginationState.ts
davecoates: Yeah maybe - know of some other implementations we can easily look at? Would rather do that than guess

eg. github uses a header to indicate pagination which does include the full URL to use https://api.github.com/repos/rails/rails/issues

js-packages/@prestojs/rest/src/getPaginationState.ts
davecoates: correct, no way around it unless the response tells us

davecoates:
> > > * Shouldn't do a full reset of paginator when changing page size
> > 
> > 
> > Why? If page size changes then the page you were on has a different meaning and could be invalid which would result in an error (even if we remove the restriction on accessing invalid pages the backend will still 404 in the case of DRF)
> 
> Can you give me an example where it would fail?
> 
> * CursorPaginator: _may_ fail if is cursor state is stored server-side and maintained between requests. If the cursor is actually just an index to fetch the next set of records from then continues to work fine
> * LimitOffsetPaginator: works fine if you change the page size. May want to round offset to a multiple of limit if changing page size
> * PageNumberPaginator: need to adjust current page number so that you are in the same spot but other than that works fine to change the page size

I read your suggestion as not altering the pagination state whereas it's now clear you mean altering the pagination to try retain the page. No problem - I went with the simple option to avoid invalid pagination states but can make it more sophisticated.

davecoates:
> > > Potential high-level changes:
> > > 
> > > * Throwing errors in InferredPaginator vs returning null
> > 
> > 
> > This essentially would ignore incorrect usage though wouldn't it? What's the motivation?
> 
> There's a few use cases:
> 
> * You're trying to call the wrong kind of paginator method on an InferredPaginator. In this case it is indeed incorrect usage, but the motivation to ignore is that it's a little more convenient to deal with null than an exception in JSX. Mostly it means that UI code that deals with all 3 pagination types (eg a generic List View) is simpler: it just hides links that don't work. (Given the infrequency at which this code is likely to happen this may be a weak justification)
> * You're calling the right kind of paginator method on an InferredPaginator, but data hasn't been returned from the server yet. In this case, again it's easier to deal with a null return value where you just eg hide the "Last" link than to deal with an exception.
> * You're calling the right kind of method but with the wrong parameter: eg a page number should never be <1. A total may be off the end though: if the last record is deleted and you click "Next" and now go to a page with no results then should the user see an error? I'd expect no, they'd just see no results.
> 
> > > I don't think we need to be so strict with limit/offset/pagesize restrictions (if data isn't back yet, or totals are out of range then we can still navigate to a page that may have empty results)
> > > You specifically can't do some things if the totals aren't back (eg. impossible to go to last page). Do you mean to just silently ignore the call entirely in those cases and in the other cases just allow it?
> > 
> > 
> > I feel silently ignoring it is bad - probably indicates your UI is not user friendly (it should be disabled until the user can interact with it). I don't have a problem allowing invalid states as that should result in feedback to the user one way or the other.
> 
> "it should be disabled until the user can interact with it" -- how can we know this with an InferredPaginator though? We don't know whether a "Last" link is even going to be possible until we get results back from the server, so we're going to have to hide the link rather than showing a loading spinner/disabled on it. If we're doing that then we should return null to make it simpler for the consumer of the API.

See my example in response to another comment. Most UI's (ie. UI you implement for a client) know the pagination type so it's a non-issue - you can render up front.

If you don't know you should render null then choose the pagination component based on the inferred type (as in my example).

> 
> Perhaps the answer is for some methods of InferredPaginator to never throw an exception and always return null whereas the underlying Paginator throws an exception on data that's never going to be valid. Data that simply isn't ready yet (as opposed to 'will _never_ be ready') would also return null

I _think_ you are saying here to just remove the errors `Cannot call XXX() until pagination type has been inferred (after call to setResponse)` and leave the errors that are thrown once pagination type is known (pagination is only know after data is ready)? I don't like it for the reason mentioned (UI just looks unresponsive until ready - should be disabled) but could compromise on this point.

davecoates:
> > > * Doesn't seem to be possible to create regular  links to do paginated navigation
> > 
> > 
> > Not sure I follow why this wouldn't work - you'd just initialise the paginator state from the URL on load.
> 
> I'm talking about generating a DOM that looks something like:
> 
> ```
>     <a href="http://mysite.com/users/?page=2&size=20">Next</a>
>     <a href="http://mysite.com/users/?page=4&size=20">Prev</a>
>     Results
>     <a href="http://mysite.com/users/?page=6&size=10">10</a>
>     <a href="http://mysite.com/users/?page=3&size=20" class="selected">20</a>
>     <a href="http://mysite.com/users/?page=2&size=40">40</a>
>     <a href="http://mysite.com/users/?page=1&size=100">100</a>
> ```
> 
> How can I get the URLs for these links? The API exposes methods to _go to_ those URLs, but it does this via `setState()` which in turn sets the URL state. There's no way for a consumer of the API to know the URL without navigating there first

OK so more generally you are talking about how to generate a state without changing the state of the paginator (ie. it's not really specific to URL's, it's just a good example). Yeah current model doesn't really lend itself to that. Any suggestions? Feels like to accomodate that nicely would necessitate a fundamental change to the design here... which I'm open to but not sure what it looks like.

Edit: The current implementation has some quirks that feel kind of wrong to me (like `syncState`) which stem from the fact that we are implementing stuff generically outside of React (eg. a class that maintains it's own state). Maybe our solution should be to ditch this and just do paginators purely as  instead.

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: (I think this is explained in your other comment that I have now responded to)

js-packages/@prestojs/rest/src/LimitOffsetPaginator.ts
davecoates: Wouldn't you just use `0`? default makes sense for page size - not so much for offset

js-packages/@prestojs/rest/src/LimitOffsetPaginator.ts
davecoates: I've removed the restriction

There's no unnecessary restrictions in `next` and `prev` though? It just checks limit is known otherwise it's not possible to go next or prev

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: I think it should clear next and previous as they are relative to cursor at old page size... it shouldn't matter because if you change page size it should refetch and update them anyway

davecoates: I've left the code that clears next/previous on change but it will now retain the current cursor on page size change

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: yep good idea

davecoates:
@levic I've pushed new commit with most these changes in it

js-packages/@prestojs/rest/src/LimitOffsetPaginator.ts
levic: Was thinking about when you pull this from a URL or uninitialised data store

davecoates: Ok done - now accepts null

django-testbed/frontend/src/views/UserListView.js
```diff
@@ -1,30 +1,84 @@
...
+            </div>
             {User.label} / {User.labelPlural}
             <Button type="primary" onClick={() => setShowCreate(true)}>
                 Add User
             </Button>
             <hr />
+            <Button onClick={() => paginator.first()}>First</Button>
+            <Button onClick={() => paginator.previous()}>Previous</Button>
+            <Button onClick={() => paginator.next()}>Next</Button>
+            <Button onClick={() => paginator.last()}>Last</Button>
```
levic, line 60: For demo purposes can you please include a mix of both buttons (onclick) and normal links (`<a href="...">`)

js-packages/@prestojs/rest/src/InferredPaginator.ts
```diff
@@ -0,0 +1,322 @@
...
+        return null;
+    }
+
+    get offset(): number | null | undefined {
+        if (this.paginator) {
+            if (!(this.paginator instanceof LimitOffsetPaginator)) {
+                return undefined;
+            }
+            return this.paginator.offset;
+        }
```
levic, line 56: This pattern is in a number of functions.

Would be simpler as either:

```js
    if (!this.paginator) return null;
    if (this.paginator instanceof LimitOffsetPaginator) return null;
    return this.paginator.limit;
```

or
```js
    if (!this.paginator || this.paginator instanceof LimitOffsetPaginator) return null;
    return this.paginator.limit;
```

(ie check for the exception and immediately abort rather than falling through and requiring nested ifs; I don't know if autoformatting is going to rewrite this though?)

js-packages/@prestojs/rest/src/InferredPaginator.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    get limit(): number | null | undefined {
+        if (this.paginator) {
+            if (!(this.paginator instanceof LimitOffsetPaginator)) {
+                return null;
+            }
+            return this.paginator.limit;
+        }
+        return null;
```
levic: Every other function returns undefined, limit and pageState return null

(although curiously the others all declare themselves as returning null | undefined but never return null, unless the underlying paginator does -- what is intended to be happening?)

js-packages/@prestojs/rest/src/InferredPaginator.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    next(): void {
+        if (this.paginator) {
+            return this.paginator.next();
+        }
+        throw new Error(
+            'Cannot call next() until pagination type has been inferred (after call to setResponse)'
+        );
```
levic: Quite minor but usually flow is easier to follow if is of the form:

- error check
- error check
- error check
- return success value

js-packages/@prestojs/rest/src/Paginator.ts
```diff
@@ -0,0 +1,54 @@
...
+     */
+    constructor([currentState, setCurrentState], [internalState, setInternalState]) {
+        this.currentState = currentState || {};
+        this.setCurrentState = setCurrentState;
+        this.internalState = internalState || {};
+        this.setInternalState = setInternalState;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    getRequestInit(options: EndpointExecuteOptions): EndpointExecuteOptions {
```
levic, line 53: Missing function header

js-packages/@prestojs/rest/src/Paginator.ts
```diff
@@ -0,0 +1,54 @@
...
+        throw new Error('Not implemented');
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    setResponse(response: Record<string, any>): void {
```
levic, line 64: Missing function header

js-packages/@prestojs/rest/src/PageNumberPaginator.ts
```diff
@@ -0,0 +1,182 @@
...
+        return {
+            query: newQuery,
+            ...options,
+        };
+    }
+
+    setResponse({ total, pageSize }: { total: number; pageSize?: number }): void {
+        this.setInternalState({ total });
+        if (pageSize && this.currentState.pageSize !== pageSize) {
+            this.setPageSize(pageSize);
```
levic, line 179: Out of curiosity, why do you need to check if the page size has changed? Can't you just set it and react will decide nothing needs to be changed (some overhead, but presumably not a lot), or does setting state trigger a rerender even if nothing changed?

js-packages/@prestojs/rest/src/PageNumberPaginator.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Return the state for the first page
+     *
+     * Does not transition state. To transition state call `last` instead.
+     */
+    lastState(): PageNumberPaginationState | null {
+        if (null == this.total || null == this.pageSize) {
+            return null;
```
levic, line 143: shouldn't this return an empty object?

js-packages/@prestojs/rest/src/CursorPaginator.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Return the state for the first page. If the next page isn't yet known (eg. results
+     * haven't yet been returned) then null will be returned.
+     *
+     * Does not transition state. To transition state call `next` instead.
+     */
+    nextState(): CursorPaginationState | null {
+        if (this.nextCursor == null) {
+            return null;
```
levic, line 86: What will happen if a consumer doesn't do a check and just calls setState on the output from this and that is null?

Is it better to throw an exception here? (Am distinguishing this from individual parts of the state, eg `total` because those are used to compose the state but you wouldn't call setState() directly with their output)

levic (changes requested):

js-packages/@prestojs/rest/src/InferredPaginator.ts
levic: (This is really minor but there's about 25-30 functions that are affected -- if you're low on time I'm happy to make these changes)

js-packages/@prestojs/rest/src/InferredPaginator.ts
davecoates: This is one where we were talking and you mentioned it and I changed these 2 immediately but forgot about others

js-packages/@prestojs/rest/src/PageNumberPaginator.ts
davecoates: setState doesn't do deep equality so will re-render as state is stored as object. I don't think it really matters though - I could have left that check out

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: The are the places I previously was throwing exceptions from that you objected to that I changed to return null... eg. so people can call them in generic components

js-packages/@prestojs/rest/src/PageNumberPaginator.ts
davecoates: It's indicating you the state isn't known (ie. you could check this is null and disable/hide the button until it's clickable)

js-packages/@prestojs/rest/src/InferredPaginator.ts
davecoates: yeah I don't know why I did it like this, I usually try to minimise nesting

(autoformatting preserves same AST always so it won't change anything)

django-testbed/frontend/src/views/UserListView.js
davecoates: Difficult right now until we add a frontend router - once #14 is available can do it easily

js-packages/@prestojs/rest/src/CursorPaginator.ts
levic: The distinction is between setters/actions & getters:
- if you say "take me to the last page" but there's no way to do that then you need to communicate that back to the caller: either a success/fail return value or throw an exception
- the same logic applies to a getter: you can either return a failure value (`null`) or throw an exception. The difference between getters & setters is that catching an exception in JSX vs checking for a null return value is very clumsy.

Something like:

```
{paginator.lastPage() &&
    <button onClick={paginator.gotoLastPage()}>Last</button>}
```

vs 

```
{(() => {
    try { return <button onClick={paginator.gotoLastPage()}>Last</button>}
    catch (err) { if (err.type instanceof WhateverException) throw err; return null; }
)()}
```


Although this doesn't really apply if there is a standard helper to ignore exception(s), something like:
```
{ignoreException(WhateverException,
    <button onClick={paginator.gotoLastPage()}>Last</button>
)}
```

davecoates: I understand all those points but don't understand what you want especially given previous comments

Can you clarify?

You've commented on `nextState` here - eg. give me the state for the 'next' page. The whole purpose of this was so you could do things like generate a URL with it which was the cases you were talking about for _not_ throwing exceptions.

davecoates: Note to self: check what currently throws; it should allow things like running of the end with next but stuff that doesn't make sense throw for (eg. going to last() on a cursor paginator)

PR #6, opened 2020-01-07
Feature/ui provider form

davecoates:


doc-site/data/final-form/Form.mdx
```diff
@@ -9,40 +9,46 @@ slug: /api/@prestojs/final-form/Form.html
...
-            value: 'data=> console.log(data)',
+            value: 'data => console.log(data)',
         },
         children: {
-            value: `
-<Form.Item field={User.fields.firstName} />
+            value: `<Form.Item field={User.fields.firstName} />
 <Form.Item field={User.fields.lastName} />
 <Form.Item field={User.fields.email} />
+<Form.Item label=" " colon={false}>
```
levic, line 32: Should label be blank here? If it's not required then why include it at all?

levic (changes requested):

doc-site/data/final-form/Form.mdx
davecoates: Without the label the button appears to the very left. With an empty label it shows in the right column in line with all the fields

levic: Smells like an issue with "" being falsy. I assume that's in the underlying library and not something in our control

davecoates: Yeah it's in antd

PR #7, opened 2020-01-08
Feature/issue18 load on demand

fanglong-alliance:


js-packages/@prestojs/ui-antd/src/__tests__/getWidgetForField.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    render(
+        <React.Suspense fallback="loading...">
+            <UnknownWidget/>
+        </React.Suspense>
+    );
+
+    const lazyElement = await screen.findByRole('spinbutton');
+    // "Property 'toBeInTheDocument' does not exist on type 'JestMatchersShape<Matchers<void, any>, Matchers<Promise<void>, any>>'". outdated def?
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
```
davecoates: should be able to fix this by adding `import '@testing-library/jest-dom/extend-expect';` to `setupTests.ts`

js-packages/@prestojs/ui-antd/src/__tests__/getWidgetForField.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const fieldArgs = {label: 'input number with a spin button'};
+    const UnknownWidget = getWidgetForField(new NumberField(fieldArgs)) as any;
+
+    render(
+        <React.Suspense fallback="loading...">
+            <UnknownWidget />
+        </React.Suspense>
+    );
+
+    const lazyElement = await screen.findByRole('spinbutton');
```
davecoates, line 17: This could just be my ignorance because I've never seen the `screen` import before - is this the idiomatic way to do this kind of thing? I thought it was more common to destructure `findByRole` from the return value of `render`?

jest.config.js
```diff
@@ -6,11 +6,7 @@ const { pathsToModuleNameMapper } = require('ts-jest/utils');
 const { compilerOptions } = require('./tsconfig.base');
 
 module.exports = {
-    setupFilesAfterEnv: [
-        'jest-dom/extend-expect',
-        'react-testing-library/cleanup-after-each',
```
davecoates, line 11: why'd you remove this?

davecoates (changes requested):

jest.config.js
fanglong-alliance: react-testing-library/cleanup-after-each is now deprecated and does not do anything. let me see if i can find the original msg

fanglong-alliance: since 9.0.0:

```
cleanup-after-each: If any of your code has import '@testing-library/react/cleanup-after-each' then you need to remove that. It's happening automatically now.
```

fanglong-alliance: deprecation message:
```

  'The module `@testing-library/react/cleanup-after-each` has been deprecated and no longer does anything (it is not needed). You no longer need to import this module and can safely remove any import or configuration which imports this module',



```

js-packages/@prestojs/ui-antd/src/__tests__/getWidgetForField.test.tsx
fanglong-alliance: i've tried render & container but both does not seem to play well with suspense (maybe you can find a way?)

the use of `screen` is taken out from react-testing-library's suspense example: 

https://github.com/kentcdodds/react-testing-library-examples/blob/master/src/__tests__/react-lazy-and-suspense.js

fanglong-alliance: (by "does not play well" i mean even after render await im still only getting the fallback "Loading..." rendered in container for some reason)

davecoates: Ok no problem - happy to go with same approach as in their examples

jest.config.js
davecoates: Ah ok cool 👍

js-packages/@prestojs/ui-antd/src/__tests__/getWidgetForField.test.tsx
fanglong-alliance: adding extend-expect to setupTests.ts doesnt solve the problem but adding it to this file do :/. i dont think tsc's seeing setupTests.ts; should maybe add it to tsconfig?

davecoates: Interesting... I tried this in `viewmodel` and it worked. Maybe some setup thing but not obvious to me what - happy to leave it for now

davecoates (approved):

PR #8, opened 2020-01-09
Feature/cache enhancements

davecoates:
These changes are to support some new hooks I'm writing. 

1) Improve ergonomics on some things (eg. fetch latest cached version of record directly)

2) Allow subscribing to any changes at all to the cache. This allows us to write hooks similar to `useSelector` in redux.

3) Allow easy retrieval of all records already cached (`getAll`). This is useful for hook that detects changes to a view model (eg. will be used to refetch a list view if any of the underlying cache data changes).


js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -9,10 +9,49 @@ import ViewModel, { PrimaryKey } from './ViewModel';
...
 class RecordPointer<T extends ViewModel> {
+    /**
+     * This is the value that was replaced by this pointer in the cache. When we actually clone
+     * the record we compare against this value - if it's the same we return the original object
+     * so that equality checks still hold. This is because we never return the RecordPointer - we
+     * only resolve the underlying value as is required. In order to determine whether or not
+     * an object is the same requires first cloning it and then comparing it. As such we do that
+     * at the time we need the cloned value, not upfront (ie. we don't do it before we store the
+     * RecordPointer as that would be lower than doing it lazily - especially if you are updating
+     * lots of records)
```
levic, line 20: Why `currentValue` rather than `currentRecord`?

I'm confused by the explanation. Up to line 15 I understand; it boils down to:

"This is the existing record. We only replace the record with a new one if field values actually change (improves performance since `isEqual()` does an object identity check first before a deep equality check)".

Everything after that is confusing (eg "we only resolve the underlying value as required") -- can we chat about it?

davecoates: What about:

> This is the existing record. We only replace a record with a new one if the fields actually change and we
> only do this when the record is accessed (if it's never accessed there's no performance and minimal
> storage costs for it).  When the record is accessed and it is a `RecordPointer` we have to clone it so
> the underlying record is returned. As part of this clone we compare `currentRecord` to the cloned
> version; if they are the same we discard the clone and return `currentRecord`.  This is purely a
> performance optimisation.

Also in case it's not clear the whole reason `RecordPointer` exists is for when you have update a set of fields (`email`, `name`, `phone`) then when records with a subset of fields are requested they get the updated values (without having to create everything upfront).

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -9,10 +9,49 @@ import ViewModel, { PrimaryKey } from './ViewModel';
...
+        this.currentValue =
+            currentValue instanceof RecordPointer ? currentValue.currentValue : currentValue;
         this.record = record;
     }
+
+    clone(fieldNames: string[]): T {
+        const cloned = this.record.clone(fieldNames);
+        // Check if the value has actually changed from what it used to be. If not we can return
+        // the old value so that equality checks still hold.
+        if (this.currentValue && cloned.isEqual(this.currentValue)) {
```
levic: Minor performance quibble: why clone and then compare? Can't you compare and then only clone if they're actually different?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
levic: That explanation is much better, but the naming of `currentValue` is confusing me even more now, especially with the fact that it's overloaded and can be a `T` _or_ a `RecordPointer<T>`.

Can we chat about this? (Specifically, how does `currentValue` differ from `record`?)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -9,10 +9,49 @@ import ViewModel, { PrimaryKey } from './ViewModel';
...
     record: T;
-    constructor(record: T) {
+    constructor(currentValue: null | undefined | T | RecordPointer<T>, record: T) {
+        this.currentValue =
+            currentValue instanceof RecordPointer ? currentValue.currentValue : currentValue;
         this.record = record;
     }
+
+    clone(fieldNames: string[]): T {
+        const cloned = this.record.clone(fieldNames);
```
davecoates, line 35: add comment indicating we clone first as field names may be subset of original record field names - we need to compare only this subset below

PR #10, opened 2020-01-09
Feature/use view model cache

davecoates:


django-testbed/frontend/src/views/UserListView.js
```diff
@@ -1,20 +1,26 @@
...
+        {
+            // refreshInterval: 1000
+        }
+    );
     if (error) {
         throw error;
     }
 
-    const records = useConnected(data);
+    const records = useViewModelCache(User, cache => (data ? cache.getList(data) : null));
```
levic: This is inconsistent with the previous pattern in `UserCreateUpdateView.js`. Either one seems fine (unless you specifically need to return `null` instead of a falsy value which is what the other one does), but should be consistent

js-packages/@prestojs/viewmodel/src/useViewModelCache.ts
```diff
@@ -0,0 +1,87 @@
...
+                lastValue.current = nextValue;
+            } catch (err) {
+                latestSubscriptionCallbackError.current = err;
+            }
+            forceRender({});
+        }
+
+        // This catches any changes to cache that occur between initial render
+        // and when the subscription is added here
+        checkForUpdates();
```
levic, line 82: Can you confirm that this is what the "should handle updates between first render and subscription" test case is checking for?

levic (changes requested):

js-packages/@prestojs/viewmodel/src/useViewModelCache.ts
davecoates: Yes

django-testbed/frontend/src/views/UserListView.js
davecoates: Done

PR #11, opened 2020-01-09
Feature/use view model change effect

davecoates:


levic:
Can we chat about this? Need a brief on what's being accomplished (mostly is unclear to me why this needs to be an effect)

django-testbed/frontend/src/views/UserListView.js
```diff
@@ -2,20 +2,23 @@ import React, { useState } from 'react';
 import { Button, Modal } from 'antd';
 import qs from 'qs';
 import { FieldFormatter } from '@prestojs/ui';
-import { useViewModelCache } from '@prestojs/viewmodel';
+import { useViewModelCache, useViewModelChangeEffect } from '@prestojs/viewmodel';
```
levic: I thought we discussed changing this to `useViewModelChange` or `useViewModelState`

js-packages/@prestojs/viewmodel/src/useViewModelChangeEffect.ts
```diff
@@ -0,0 +1,174 @@
...
+};
+
+type UpdatedDiff<T extends ViewModel> = {
+    previous: T[];
+    updated: T[];
+};
+
+type DeletedDiff<T extends ViewModel> = {
+    previous: T[];
+    deleted: T[];
```
levic: Is it going to be annoying to use if added/updated/deleted are different?

If the fields are the same `previousState` / `newState` regardless of whether it's an Added/Updated/Deleted operation then you don't need a `switch`/`if` based on the event type -- or is that something you're actively wanting to be treated differently?

(or the "light" version of this: just Added/Updated be the same thing since they're often treated the same way)

js-packages/@prestojs/viewmodel/src/useViewModelChangeEffect.ts
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @param records The records to monitor for changes. This should usually
+ * come from a `useViewModelCache` or another hook that returns data live
+ * from he cache. If a falsy value is passed then no changes will be detected.
+ * Use this to opt out of checks (for example when a fetch is in progress).
+ * @param run The function to call when `records` changes. The first parameter
+ * will be one of 'ADD', 'UPDATE' or 'DELETE' depending on the change. If
+ * `options.includeDiff` is true then the second parameter will be an object
+ * describing the change with the shape:
+ *
```
levic: `includeDiff` is not set by default since it involves extra processing? Does this need to be mentioned so that devs don't do it out of habit or is it actually not that expensive?

js-packages/@prestojs/viewmodel/src/useViewModelChangeEffect.ts
```diff
@@ -0,0 +1,174 @@
...
+ * ```jsx
+ * export default function UserListView() {
+ *   const { data, revalidate, isValidating } = useEndpoint(User.endpoints.list);
+ *   // Refetch data whenever underlying cache changes
+ *   const allRecords = useViewModelCache(User, cache => cache.getAll(fieldList));
+ *   // Note that we pass false while data is being fetched from the backend
+ *   useViewModelChangeEffect(!isValidating && allRecords, revalidate);
+ *   return <ListView records={data} />;
+ * }
```
levic: After discussion,

- On the first render
    - useEndpoint() triggers an ajax call; isValidating=true
    - allRecords is empty since nothing is in the cache yet
    - useVieModelChangeEffect does nothing
- Once the ajax call has returned
    - useEndpoint() populates the cache; isValidating=false
    - allRecords is now the full list of fetched records from useEndpoint()
    - useViewModelChangeEffect now is passed in a list of records but because it was a transition from falsy does not trigger a revalidate()

levic (changes requested):

js-packages/@prestojs/viewmodel/src/useViewModelChangeEffect.ts
davecoates: they have to be different as they can all happen (the diff is a union of all these types - see `Diff` below)

js-packages/@prestojs/viewmodel/src/useViewModelChangeEffect.ts
davecoates: yeah it can be expensive, i can add a note

js-packages/@prestojs/viewmodel/src/useViewModelChangeEffect.ts
levic: Can you add a comment somewhere that

- usually you don't want multiple useViewModelChangeEffects on the exact same subset of data as you will trigger redundant ajax queries.
- if you a record is updated & saved elsewhere, then useViewModelCache will return new data which trigger useViewModelChangeEffect. We still want useViewModelChangeEffect () to trigger the useEndpoint() revalidate because that record update may have changed the order of records or caused it to [no longer] appear in a filtered list of data

js-packages/@prestojs/util/src/useListChangeObserver.ts
```diff
@@ -76,7 +76,13 @@ type Change<T> = {
  *   const { data, revalidate, isValidating } = useEndpoint(User.endpoints.list);
  *   // Refetch data whenever underlying cache changes
  *   const allRecords = useViewModelCache(User, cache => cache.getAll(fieldList));
- *   // Note that we pass false while data is being fetched from the backend
+ *   // NOTE: Usually you don't want multiple useViewModelChangeEffects on the exact same
+ *   // subset of data as you will trigger redundant ajax queries.
+ *   // if you a record is updated & saved elsewhere, then useViewModelCache will
```
levic: "if you a record"?

(This might be my only comment but I promise I read the whole thing!)

js-packages/@prestojs/viewmodel/src/useViewModelChangeEffect.ts
levic: I presume there's nothing for this because it's not relevant, with separate useChangeObserver and useListChangeObserver classe?

levic (approved):
lgtm apart from the 2 outstanding comments

js-packages/@prestojs/util/src/useListChangeObserver.ts
davecoates: oops! fixed

davecoates:
> I presume there's nothing for this because it's not relevant, with separate useChangeObserver and useListChangeObserver classe?

yep and the implementation has changed in `useListChangeObserver` to be more robust and includes the details of the `change` automatically in the first argument to `onChange`. The old implementation relied on only 1 operation happening at a time which meant if you had an addition and delete then it wouldn't detect them (which worked in the context of current caching implementation which doesn't batch changes but doesn't work in general). The new implementation doesn't have this limitation and as such necessarily always knows what's changed.

PR #12, opened 2020-01-10
feat: frontend FilterSet

fanglong-alliance:
affects: django-testbed, @prestojs/final-form, @prestojs/viewmodel

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/FilterSet.ts
+++ b/js-packages/@prestojs/viewmodel/src/FilterSet.ts
@@ -1,5 +1,4 @@
 import FieldBinder from './FieldBinder';
-import ViewModel from './ViewModel';
 
 /**
  * Base Filterset class for filters to be applied to models. Shares commonality with ViewModels in sense that FilterSets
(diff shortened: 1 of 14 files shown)


js-packages/@prestojs/viewmodel/src/FilterSet.ts
```diff
@@ -0,0 +1,49 @@
...
+ *         // label is optional; will be generated as 'Last name'
+ *         lastName: new CharField(),
+ *     };
+ * }
+ * ```
+ *
+ * @extract-docs
+ */
+export default class FilterSet extends FieldBinder {
+    static _model: typeof ViewModel; // TODO - review whether we do want this; guess depends on whether filtering-on-related-model is a thing?
```
davecoates: I don't think we need this - or at least lets leave it out until we discover a need for it

js-packages/@prestojs/viewmodel/src/FilterSet.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    [fieldName: string]: any;
+
+    public get _filterSet(): typeof FilterSet {
+        return Object.getPrototypeOf(this).constructor;
+    }
+
+    constructor() {
+        super();
+        if (!this._filterSet.fields || !this._filterSet.__boundFields.has(this._filterSet)) {
```
davecoates: I think this may as well be an error - if you instantiate a filterset it's kind of meaningless right now

davecoates (changes requested):

js-packages/@prestojs/viewmodel/src/FilterSet.ts
fanglong-alliance: gp

js-packages/@prestojs/viewmodel/src/FilterSet.ts
```diff
@@ -26,21 +25,15 @@ import ViewModel from './ViewModel';
...
         super();
-        if (!this._filterSet.fields || !this._filterSet.__boundFields.has(this._filterSet)) {
-            throw new Error(
-                `Class ${this._filterSet.name} has not been defined correctly. Make sure field definitions are set on the '_fields' property and not 'fields'.`
-            );
-        }
+        throw new Error(
+            `${
+                Object.getPrototypeOf(this).constructor.name
+            }: FilterSet serves as a collection of fields and should not be instantized.`
```
davecoates: instantized => instantiated

Maybe clarify the static nature, eg. `FilterSet serves as a static collection of fields and should not be instantiated`

davecoates (changes requested):

davecoates (approved):

PR #13, opened 2020-01-13
feat: partial fields via url on serializer

fanglong-alliance:
affects: django-testbed


django-testbed/django-root/xenopus_frog/views.py
```diff
@@ -7,14 +7,57 @@
 from xenopus_frog.models import User
 
 
+class BunniesSerializer(ModelSerializer):
```
davecoates: 🐰

davecoates: Is there a reason we wouldn't put this in the library?

django-testbed/django-root/xenopus_frog/views.py
```diff
@@ -7,14 +7,57 @@
...
+    """
+    Regulates fields exposed by default & as requested.
+
+    Construct URLs like /?include_fields=first_name&include_fields=email to opt-in into fields.
```
davecoates: I think this will be too verbose - URL will be massive in lots of cases

I think `include_fields=first-name,email` will be nicer

@levic thoughts?

django-testbed/django-root/xenopus_frog/views.py
davecoates: I think maybe we shouldn't extend anything? just use as a mixin that assumes `self.fields` and `self.Meta` exists

django-testbed/django-root/xenopus_frog/views.py
```diff
@@ -7,14 +7,57 @@
...
+
+    If any of fields specified to be included does not belong to opt_in_only_fields, the serializer
+    will perform in "strict" mode that ONLY fields specced are included.
+    """
+
+    def __init__(self, *args, **kwargs):
+        result = super().__init__(*args, **kwargs)
+
+        # only if context.request is available (ie. not during inspection etc)
+        if hasattr(self, "context") and "request" in self.context:
```
davecoates: djrad supported passing in partial fields in context as well as in request. this was useful when serializing stuff manually and was used on at least one project (foodies). we should support here as well

django-testbed/django-root/xenopus_frog/views.py
```diff
@@ -7,14 +7,57 @@
...
+            heavy_fields = set(getattr(self.Meta, "opt_in_only_fields", []))
+
+            if "include_fields" in self.context["request"].GET:
```
davecoates: [query_params](https://www.django-rest-framework.org/api-guide/requests/#query_params) rather than `GET`

django-testbed/django-root/xenopus_frog/views.py
```diff
@@ -7,14 +7,57 @@
...
+                include_fields = dict(self.context["request"].GET)["include_fields"]
+                if not type(include_fields) == list:
```
davecoates: I'm guessing the type here is out of our control so should be safe to assume is only ever a list.. but to be paranoid probably `if not isinstance(include_fields, (list, tuple))` is better

davecoates (changes requested):
As discussed move to a core library package

Needs test cases

django-testbed/django-root/xenopus_frog/views.py
fanglong-alliance: as we're doing the splitting (comma separated) this no longer matters - removing type casting

fanglong-alliance:
fyi: a little more changes will be up in next MR (incl renaming to presto_drf, relocate softlink, move to mixins.py etc)

python-packages/presto_viewmodels_drf/presto_viewmodels_drf/serializers.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, *args, **kwargs):
+        result = super().__init__(*args, **kwargs)
+
+        # only if context.request is available (ie. not during inspection etc)
+        if hasattr(self, "context") and (
+            "request" in self.context or "include_fields" in self.context
+        ):
+            heavy_fields = set(getattr(self.Meta, "opt_in_only_fields", []))
+
+            if "include_fields" in self.context["request"].query_params:
```
davecoates: this will break if no 'request'

(which highlights you are missing a test for `context["include_fields"]`)

python-packages/presto_viewmodels_drf/presto_viewmodels_drf/serializers.py
```diff
@@ -0,0 +1,46 @@
+class OptinFieldsSerializerMixin:
```
davecoates: I'm not sure on naming. Is it unnecessarily verbose? Perhaps we should just go with `OptInFieldsMixin`? The package name indicates it's for serializer

(either way should be `OptIn` rather than `Optin`)

python-packages/presto_viewmodels_drf/presto_viewmodels_drf/serializers.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    Regulates fields exposed by default & as requested.
+
+    Construct URLs like /?include_fields=first_name,email to opt-in into fields.
+
+    By default, all "opt_in_only_fields" are excluded from serializer, unless specifically opted-in.
+
+    If any of fields specified to be included does not belong to opt_in_only_fields, the serializer
+    will perform in "strict" mode that ONLY fields specced are included.
```
davecoates: This sentence is confusing

Isn't it just that if you specify any fields then only those fields are included (ie. it's irrelevant what it's opt_in_only_fields - that's only used if you don't specify  anything)?

davecoates: Perhaps just

> If `include_fields` is specified then only those fields are serialized.

python-packages/presto_viewmodels_drf/presto_viewmodels_drf/serializers.py
```diff
@@ -0,0 +1,46 @@
...
+                else:
+                    include_fields = set(self.context["include_fields"])
+            else:
+                include_fields = set(self.Meta.fields).difference(heavy_fields)
+
+            strict = not heavy_fields.issuperset(include_fields)
+
+            # in strict mode, only fields inside include_fields are included
+            # otherwise, all heavy_fields not inside include_fields gets ejected
+            if strict:
```
davecoates: ah... I see what you meant in the comment now

I don't think this works though - how would you only request a field that's in `heavy_fields`? You'd end up with all the non-heavy fields too

davecoates (changes requested):

python-packages/presto_viewmodels_drf/presto_viewmodels_drf/serializers.py
fanglong-alliance: 👏

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,65 @@
...
+    /?include_fields=first_name,email&opt_in_fields=gait_recognition_prediction
+
+    1. By default, all "fields" defined in serializer, minus those listed in "opt_in_fields" would be returned.
+    2. If "include_fields" is supplied, only fields requested this way would be returned.
+    3. If "opt_in_fields" is supplied, fields requested this way PLUS fields from #1 or #2 would be returned.
+
+    Pinned fields are always returned. (currently pk only if pk exists)
+    """
+
+    def get_pinned_fields(self):
```
davecoates: can you add header comment that will show up in docs + return type?

davecoates (changes requested):
Looks good - just missing that one method header comment

python-packages/presto_drf/presto_drf/mixins.py
fanglong-alliance: done

fanglong-alliance:
rebasing

davecoates (approved):

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+        Override on serializer to customize.
+
+        Currently, the only pinned field is pk of model (with ModelSerializer).
+
+        :return: [] - a list of pinned fields.
+        """
+
+        try:
+            return [self.Meta.model._meta.pk.name]
+        except Exception:
```
levic: - that seems a very broad exception to catch
- every django model has to have a PK, so what exceptions can legitimately be thrown here?

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+    /?include_fields=first_name,email&opt_in_fields=gait_recognition_prediction
+
+    1. By default, all "fields" defined in serializer, minus those listed in "opt_in_fields" would be returned.
+    2. If "include_fields" is supplied, only fields requested this way would be returned.
+    3. If "opt_in_fields" is supplied, fields requested this way PLUS fields from #1 or #2 would be returned.
+
+    Pinned fields are always returned. (currently pk only if pk exists)
+    """
+
+    def get_pinned_fields(self) -> List:
```
levic: Should be `-> Collection[str]` (you can iterate over the result and do `in`) -- see summary at https://alliancesoftware.atlassian.net/wiki/spaces/TEC/pages/230817832/Type+Hinting+typing

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+
+        try:
+            return [self.Meta.model._meta.pk.name]
+        except Exception:
+            pass
+        return []
+
+    def __init__(self, *args, **kwargs):
+        result = super().__init__(*args, **kwargs)
```
levic: Why is there a return value on `__init__()`?

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+
+        if not hasattr(
+            self, "context"
+        ):  # serializer invoked without context - inspection?
+            return result
```
levic: Is this the code autoformatter trying to keep the lines short? becomes harder to read.
```py
       if not hasattr(self, "context"):
            # serializer invoked without context - probably an inspection
            return result
```

levic: (or)

```py
try:
    self.context
except AttributeError:
    # serializer invoked without context - probably an inspection
    return result
```

python-packages/presto_drf/presto_drf/tests/test_optin_field_mixin.py
```diff
@@ -0,0 +1,105 @@
...
+from xenopus_frog.models import User
+
+from ..mixins import SerializerOptInFieldsMixin
+
+
+class UserSerializer(SerializerOptInFieldsMixin, ModelSerializer):
+    class Meta:
+        model = User
+        fields = ("id", "first_name", "last_name", "email", "activated_at", "is_staff")
+        opt_in_only_fields = ["activated_at", "is_staff"]
```
levic: Why is `opt_in_only_field` a list but `field` is a tuple? seems a bit random.

.flake8.cfg
```diff
@@ -1,8 +1,8 @@
 [flake8]
-ignore = E501, E231
+ignore = E501, E231, E203, W503
```
levic, line 2: Is E203  ("whitespace before :") necessary? searching this changeset it never happens, and I don't know of a time when it would be ok

python-packages/presto_drf/presto_drf/mixins.py
levic: Edit: this is a serializer, not a model (so there may be no PK). Can we narrow down the exception then?

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
+from typing import List
+
+
+class SerializerOptInFieldsMixin:
+    """
+    Regulates fields exposed by default & as requested based on query parameters or context.
+
+    Pass 'include_fields' / 'opt_in_fields' thru query params or context to use; multiple fields are separated by comma eg,
+    /?include_fields=first_name,email&opt_in_fields=gait_recognition_prediction
```
levic, line 10: I think we should be able to handle both comma separated fields and multiple field names for include_fields & opt_in_fields (depending on the context, either may be more convenient for the frontend to pass), documentation should include this

`?include_fields=first_name,last_name,email`
and
`?include_fields=first_name&include_fields=last_name`
and even mixed
`?include_fields=first_name&include_fields=last_name,email`

levic: There doesn't seem to be any documentation (outside header comments).

Even if you don't fill it in, please add a placeholder for SerializerOptInFieldsMixin documentation

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+            or []
+        )
+
+        for f in list(self.fields.keys()):
+            if (
+                f not in pinned_fields
+                and f not in fields_to_include
+                and f not in opt_in_fields_to_include
+            ):
+                del self.fields[f]
```
levic, line 86: How does this work? Doesn't the metaclass set`self.fields` as a static class property? If you delete fields from it then the next request will have a missing field

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+            (
+                "request" in self.context
+                and "include_fields" in self.context["request"].query_params
+                and self.context["request"].query_params["include_fields"]
+            )
+            or ("include_fields" in self.context and self.context["include_fields"])
+            or set(self.Meta.fields).difference(set(fields_to_exclude))
+        )
+
+        if type(fields_to_include) == str:
```
levic: shouldn't use equality for types, should use `isinstance` (python allows str subclassing)

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+        fields_to_exclude = getattr(self.Meta, "opt_in_only_fields", [])
+
+        fields_to_include = (
+            (
+                "request" in self.context
+                and "include_fields" in self.context["request"].query_params
+                and self.context["request"].query_params["include_fields"]
+            )
+            or ("include_fields" in self.context and self.context["include_fields"])
+            or set(self.Meta.fields).difference(set(fields_to_exclude))
```
levic, line 57: Why using `self.Meta.fields` and not `self.fields`? Later on you use `self.fields.keys()`, seems inconsistent

python-packages/presto_drf/presto_drf/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def __init__(self, *args, **kwargs):
+        result = super().__init__(*args, **kwargs)
+
+        if not hasattr(
+            self, "context"
+        ):  # serializer invoked without context - inspection?
+            return result
+
+        pinned_fields = self.get_pinned_fields()
```
levic, line 46: Here is a simplified version (at least I see it as simpler, esp re handling of things that may (not) be present, eg query_params)

```py
try:
	query_params = self.context['request'].query_params
except KeyError, AttributeError:
	query_params = {}

def parse_includes(spec: Union[str, Iterable[str]]) -> Set[str]:
    # turns query params with a source of 
    #  ?include_fields=first_name&include_fields=last_name,email
    # into ['first_name', 'last_name', 'email']
    if isinstance(spec, str): spec = [spec]
	return set(flatten(f.split(',') for f in spec)))

try:
	include_fields = parse_includes(query_params['include_fields'])
except KeyError:
	include_fields = set(self.Meta.fields) - getattr(self.Meta, "opt_in_only_fields", set())

try:
	opt_in_fields = parse_includes(query_params['opt_in_fields'])
except KeyError:
	opt_in_fields = set()

final_fields = set(self.get_pinned_fields()) + include_fields + opt_in_fields

# this ignores misspelled fields; can iterate through final_fields to build up self.fields 
# if you want to throw an error instead
self.fields = {
	name: field
	for name, field
	in self.fields
	if name in final_fields
}
```

(Haven't tested this, in particular `parse_includes` may be wrong: my recollection of the behaviour of django's query parser is:
-  `?a=x` => `query_params['a'] == 'x'`
- `?a=x&a=y` => `query_params['a'] == ['x', 'y']`

levic (changes requested):

.flake8.cfg
fanglong-alliance:
```
...
Since E203 is not PEP 8 compliant, you should tell Flake8 to ignore these warnings.
```

fanglong-alliance: (^ black doc; I think not PEP8 complaint's a pretty good reason to do as advised)

python-packages/presto_drf/presto_drf/mixins.py
fanglong-alliance: not restricted to ModelSerializer, hence you might not have a model at all. 
updated.

python-packages/presto_drf/presto_drf/mixins.py
fanglong-alliance: i thought we're generating docs from header comments?

fanglong-alliance: on the first one: both of not-mixed-usages are supported already.  (added testcase to cover for that)

do we really want to support the mixed usage? why?

python-packages/presto_drf/presto_drf/mixins.py
fanglong-alliance: we want the original, "official" fields, not possibly-mutated-elsewhere ones to do the comparison for what to include, but we want to only remove fields present - does this make sense?

python-packages/presto_drf/presto_drf/mixins.py
fanglong-alliance: not sure i understand here. 

```
    def test_only_include_fields_specified_are_returned_single(self):
        context = {"request": MockRequest({"include_fields": "email"})}
        serializer = UserSerializer(context=context)
        self.assertTrue("id" in serializer.fields)
        self.assertTrue("email" in serializer.fields)
        self.assertTrue("first_name" not in serializer.fields)
        self.assertTrue("activated_at" not in serializer.fields)
        self.assertTrue("is_staff" not in serializer.fields)

        context = {"request": MockRequest({"include_fields": "first_name"})}
        serializer = UserSerializer(context=context)
        self.assertTrue("id" in serializer.fields)
        self.assertTrue("email" not in serializer.fields)
        self.assertTrue("first_name" in serializer.fields)
        self.assertTrue("activated_at" not in serializer.fields)
        self.assertTrue("is_staff" not in serializer.fields)
```

ie, why would serializers get reused between requests?

davecoates: `self.fields` is not a static property, here's the code:

```python
    @cached_property
    def fields(self):
        """
        A dictionary of {field_name: field_instance}.
        """
        # `fields` is evaluated lazily. We do this to ensure that we don't
        # have issues importing modules that use ModelSerializers as fields,
        # even if Django's app-loading stage has not yet run.
        fields = BindingDict(self)
        for key, value in self.get_fields().items():
            fields[key] = value
        return fields
```

mutating this is the only way to change what's used for only that serializer instance (ok it's not - you can modify `to_representation` but less efficient as still generates all the data first before you remove it)

python-packages/presto_drf/presto_drf/mixins.py
levic: > do we really want to support the mixed usage? why?

If you've got different parts of the system that are being composed then they each might want to include their own fields. If the first one uses the shorthand syntax (`?include_fields=foo,bar`) in a query string then you can't easily add another field to the list of fields to include unless you parse the query string and recreate it.

python-packages/presto_drf/presto_drf/mixins.py
levic: The logic as implemented is the union of the original set of fields and the current set of fields, that's the intent?

I assume self.fields is set in `super()`?

levic: Edit: I've seen the `fields` implementation: if `fields` has been mutated elsewhere then shouldn't you be respecting the fact that something else has modified it?

python-packages/presto_drf/presto_drf/mixins.py
levic: > self.fields is not a static property

I didn't realise this; my comment can be ignored

levic (approved):

PR #14, opened 2020-01-14
Feature/use url query state

davecoates:
This is the same as in djrad apart from not relying on react-router anymore. There's 2 new options to make this possible:

`location` - the current location (previously hardcoded to react-router)
`replaceUrl` - function to replace current url (previously hardcoded to react-router history)

This makes the hook generic to any router (examples provided for react-router & navi)


levic:
Is install-state.gz supposed to be committed? Is [ignored here](https://github.com/yarnpkg/berry/blob/master/.gitignore) and [here](https://github.com/babel/babel/blob/master/.gitignore) so I assume not?

django-testbed/frontend/src/useEndpoint.js
```diff
@@ -9,8 +9,8 @@ import useSWR from 'swr';
  * @return Object Same values as returned by useSWR with the addition of `execute` which
  * can be used to execute the action directly, optionally with new arguments.
  */
-export default function useEndpoint(action, args, config) {
-    const paginator = usePaginator(action && action.getPaginatorClass());
+export default function useEndpoint(action, args, { paginationStatePair, ...config } = {}) {
```
levic, line 12: I was wondering about the consistency of `paginationStatePair` vs `paginationState` vs `paginationUseState` etc. [Here](https://github.com/prestojs/prestojs/pull/30/files#diff-ac3463492dddcc2cdcda76fde5083514L38) we you didn't even name it.

Can you confirm that this is consistent with how this has been named elsewhere? Are there any other project you know of (esp in React itself) that have a convention for how to name this?

django-testbed/frontend/src/views/UserFilterForm.js
```diff
@@ -11,7 +11,7 @@ function ClearButton() {
     return (
         <Button
             onClick={() => {
-                api.reset();
+                api.reset({});
```
levic, line 14: Is the fact that you had to do that a deficiency in useForm()?

js-packages/@prestojs/routing/src/useUrlQueryState.ts
```diff
@@ -0,0 +1,447 @@
...
+};
+
+function identity<T>(a: T): T {
+    return a;
+}
+
+// TODO: Should the default parse actually do something special here? eg. if it's
+// 'true' or 'false' transform to bool? if it looks like a number convert to number?
+// Date handling?
+// TODO: Should stringify have any special cases? Dates?
```
levic: To add to this list: arrays, nested objects?

js-packages/@prestojs/routing/src/useUrlQueryState.ts
```diff
@@ -0,0 +1,447 @@
...
+): [
+    /**
+     * The current query state
+     *
+     * If `options.prefix` is specified this contains only query param keys that
+     * start with the prefix otherwise it contains _all_ query params.
+     */
+    Record<string, any>,
+    /**
+     * State transition function. Accepts either the state object to transition
```
levic, line 309: Is this just "State transition function, the same as useState() with the following differences:"

js-packages/@prestojs/routing/src/useUrlQueryState.ts
```diff
@@ -0,0 +1,447 @@
...
+        replaceUrl = typeof window != 'undefined' &&
+            ((url: string): void => window.history.pushState(null, '', url)),
+    } = options;
+    if (!location || !replaceUrl) {
+        throw new Error('The url and replaceUrl options must be provided');
+    }
+    const { search, pathname } = location;
+
+    // This effect is used to update URL to include any missing keys that are
+    // specified in initialState. It only runs once - any subsequent changes to
```
levic, line 376: If `initialState` is `{foo:undefined}` or `{foo:null}` then what happens?

js-packages/@prestojs/routing/src/useUrlQueryState.ts
```diff
@@ -0,0 +1,447 @@
...
+     * eg. Calling `setUrlState({ count: 1, name: 'test' });` would call `parse`
+     * twice, once with `1, 'count'` and once with `'test', name`.
+     */
+    parse?: Parse;
+    /**
+     * A function used to stringify a value to be stored in the URL. If not specified the value's toString() method will be called. This could be useful if needing to store a complex object in the query string.
+     *
+     * It is passed 2 arguments; the value to stringify and the name of the parameter.
+     */
+    stringify?: Stringify;
```
levic: Suggestions:
- allow for specification of the field name in the options; this would allow you to use utility functions easier (you could do it with a factory function but this is more straightforward). It may still be helpful to pass the field name, as the 2nd param; functions can ignore it if they want
- seems easier to specify stringify + parse together, would keep encoding/decoding together & also easier use of utility functions (see the date example below)


```
// simplified example (no error handling on most of these)
params: {
  // decoding:
  count: x => int(x),

  // boolean decoding + encoding
  // a utility function might also handle null/undefined and strings like
  // 'yes'/'false'/'no' etc
  isActive: [
    x => x == 'true' || x > 0 : true : false,
    x => x ? '1' : '0',
  ],

  // more complicated encoding / decoding:
  createdAt: [
    x => x != '' ? moment(x) : null,
    x => x ? x.format('YYYY-mm-dd') : null,
  ],

  // using a utility function [not a big fan of the naming convention for this
  // function, but can't think of anything shorter/clearer]
  createdAt : urlParamOptionalMoment,
  // or with params
  createdAt : urlParamMoment({ optional: true }),

  // wildcard for anything else that doesn't match (not sure about `'*'` -- maybe
  // a symbol/`null`/class constant?)
  // the fact that the 2nd param is the field name allows disambiguation of the function
  // feels a little clumsy that encoding/decoding are separate through. not sure if this can
  // be made better without making the use case of passing only a decode function harder
  // (the 'count' example above)
  '*': [myDecodingFunc, myEncodingFunc],
}
```

levic: At this level complexity I start to wonder whether typescript can be used to specify the expected param types or whether that's overcomplicating things.

js-packages/@prestojs/routing/src/useUrlQueryState.ts
```diff
@@ -0,0 +1,447 @@
...
+     * If not specified all query params are sync'd. To make this work nicely
+     * with other usages of the hook on the same page consider setting
+     * `controlledKeys`
+     */
+    prefix?: string;
+    /**
+     * If specified only these keys will be synced to and read from the request
+     * URL. Any other keys will be ignored.
+     */
+    controlledKeys?: Array<string>;
```
levic: If `controlledKeys == true` then could default to all the keys in the `params` list above (and throw an exception if this is combined with a wildcard/fallback key)

levic (changes requested):

django-testbed/frontend/src/useEndpoint.js
davecoates: I couldn't find any naming convention externally - almost always destructured. It's called `currentStatePair` in the pagination branch but that's in the context of paginator - here there's no such context so either `paginationStatePair` or `paginationCurrentStatePair`

Type definitions in React source are an unnamed tuple and I couldn't find any examples of naming it without destructuring

django-testbed/frontend/src/views/UserFilterForm.js
davecoates: not a deficiency, just how it works

with no parameters it resets it to values form was initialised with (not what I want here - should clear form entirely given it's initialised with values from the URL)

js-packages/@prestojs/routing/src/useUrlQueryState.ts
davecoates: for null url will have ?foo= 

for undefined won't appear

js-packages/@prestojs/routing/src/useUrlQueryState.ts
davecoates: yes

js-packages/@prestojs/routing/src/useUrlQueryState.ts
davecoates: So remove parse/stringify and replace with `params`? Seems good to me

Are you suggesting the two options for each parameter are either a tuple of encode/decode OR a single function which is to decode (ie. there's a shortcut for decode but not encode?)

davecoates:
I've updated this with `params` option (removed `parse`/`stringify`)

js-packages/@prestojs/routing/src/useUrlQueryState.ts
levic: Can you make that more explicit please

js-packages/@prestojs/routing/src/useUrlQueryState.ts
levic: What do other libraries do?

Intuitively I would expect that `null` would remove the param.

`undefined` I would think would _maybe_ remove it, but I would regard that as ambiguous (is the desired target state undefined, or is the desired target state _set to_ undefined?)

js-packages/@prestojs/routing/src/useUrlQueryState.ts
levic: > Are you suggesting ...

Yes

Am not particularly strongly in favour of anything here, just trying to think through what's going to make the common use cases easier to deal with so if you come up with places where this doesn't work then feel free to change it

js-packages/@prestojs/routing/src/useUrlQueryState.ts
davecoates: Well in this case we are doing exactly what another library does...

```
qs.stringify({ test: null, ok: 5 })
```

> "test=&ok=5"

In general undefined means it's not set, `null` is a valid value

js-packages/@prestojs/routing/src/useUrlQueryState.ts
davecoates: What do you suggest? the latest version since your original comment already includes the line 

> Passing the function returned by `useState` is compatible with this.

django-testbed/frontend/src/views/UserFilterForm.js
levic: Shouldn't the URL only add/override default values?

Consider a report page that defaults to dateStart=today(), dateEnd='' (infinity)

What would you expect of http://foo.com/my-page?dateEnd=2020-04-20 ?

levic: (Also, what you would expect here may conflict with what you would expect from setting dateStart=null in code)

js-packages/@prestojs/rest/src/LimitOffsetPaginator.ts
```diff
@@ -19,11 +19,19 @@ export default class LimitOffsetPaginator extends Paginator<
     }
 
     get limit(): number | null {
-        return this.currentState.limit ?? null;
+        const limit = this.currentState.limit ?? null;
+        if (typeof limit === 'string') {
+            return Number(limit);
+        }
```
levic, line 25: Minor style: when I have multiple type changes like that I find it reads more linearly as:

```
if typeof(limit === 'string') {
    limit = Number(limit);
}
return limit;
```

since it minimises the number of return points in the function (in this case it doesn't really matter since the lines are next to each other, but if more code gets added in after the if {} then it starts to become harder to follow) [appears in a few other places too]

levic: (Edit: on the other hand exceptions/errors/guard clauses usually _are_ points of early return)

js-packages/@prestojs/routing/src/useUrlQueryState.ts
```diff
@@ -0,0 +1,483 @@
...
+     *
+     * The decode function is used to parse query values to transform them to a certain type or shape.
+     *
+     * The encode function is used stringify a value to be stored in the URL. When using the
+     * single function form (just the decode function) then the encode function is the equivalent
+     * of calling the toString() method on the value.
+     *
+     * Both encode and decode are passed 2 arguments; the value to transform and the name of the parameter.
+     *
+     * A special key '*' can be set to define the fallback behaviour for all keys not explicitly defined.
```
levic, line 30: Is there a way of telling typescript that `'*'` is the only valid string value? (see ParamsEncodeDecode above)

If not, maybe a Symbol is better? (or `null`?)

js-packages/@prestojs/routing/src/useUrlQueryState.ts
```diff
@@ -0,0 +1,483 @@
...
+     * are the rules that are applied when determining what to keep in the URL:
+     *
+     * If `options.controlledKeys` is not specified then any query param keys
+     * that match the specified list will be removed if not included in the new
+     * state object. Any keys not in the specified list will be retained.
+     *
+     * If `options.controlledKeys` is not specified and `options.prefix` is specified
+     * then any query param keys that start with the prefix will be removed if
+     * not included in the new state object. Any keys not starting with `prefix` will
+     * be retained.
```
levic, line 328: It talks about what happens if `controlledKeys` is not present but I can't find anywhere that says what it actually does (or that `true` means "all keys")

js-packages/@prestojs/routing/src/useUrlQueryState.ts
```diff
@@ -58,6 +65,25 @@ const pickWithoutKeys = (
     withoutKeys: Array<string>
 ): Record<string, any> => pickBy(obj, (value: any, key: string) => !withoutKeys.includes(key));
 
+function encode(key: string, value: any, params: ParamsEncodeDecode): string {
+    const p = params[key] || params[WILDCARD];
+    if (!p || !Array.isArray(p)) {
+        return value;
```
levic: The comments in `Options` say `toString(value)` but this just returns `value`

I think is then implicitly converted to a string by `qs`, but would be better to be explicit

js-packages/@prestojs/routing/src/useUrlQueryState.ts
levic: Ah I missed that the change was at the end of the comment block, I thought you hadn't changed anything.

> Passing the function returned by

Is that a typo? Isn't it the transition function passed _into_ `useState()` rather than the state setting function returned from `useState()` that is compatible?


This makes sense to me:

```
    /**
     * State transition function. Accepts either the state object to transition
     * to OR a transition function.
     *
     * A transition function should accept the current state and return the new
     * state (similar to a `useState` transition function).
     *
     * Unlike `useState` the returned state always replaces the current state - it is not merged
     * except with the caveats noted below.
```

(Feel free to reword if I'm wrong or you think could be clearer)

levic (changes requested):

django-testbed/frontend/src/views/UserFilterForm.js
davecoates: If you have default values that you want to retain in the form itself then the implementation for the button will be different yes but in this testbed code there's not

js-packages/@prestojs/routing/src/useUrlQueryState.ts
davecoates: It's not correct - `useState` always replaces the state it never merges same as this

The only thing you (optionally) pass to `useState` is the initial state. It returns the state and the function to call to change state. So what I wrote was correct in that it's the return value of `useState` you pass to this

js-packages/@prestojs/routing/src/useUrlQueryState.ts
davecoates: hmm but it's not the only valid string value? every key in params is a string...

you can definitely type things with a literal but don't see how it helps here

js-packages/@prestojs/routing/src/useUrlQueryState.ts
davecoates: it's described in the header doc for `controlledKeys`. It didn't say what `true` meant - I've added this now

PR #18, opened 2020-01-15
fix: fix buggy choices formatter & test

fanglong-alliance:
affects: django-testbed, @prestojs/ui


django-testbed/django-root/xenopus_frog/models.py
```diff
@@ -33,6 +33,15 @@ class User(GenericUserProfile, AbstractEmailUser):
...
     email = models.EmailField(unique=True)
 
+    CHOICES_AGE = [
+        [1, "One"],
+        [2, "OneOne"],
+        [3, "OneOneOne"],
+        [4, "OneOneOneOne"],
+    ]
+
+    age = models.IntegerField(null=True, blank=True, choices=CHOICES_AGE)
```
davecoates: Long term it will be nicer to keep these models representative of something that might exist in practice... can you change this to something like contact_type with 'Lead', 'Contact', 'Supplier', 'Client' that's vaguely real?

davecoates (commented):
Otherwise looks good :+1:

django-testbed/django-root/xenopus_frog/models.py
fanglong-alliance: updated; also added E231 ignore to flake8 (its hitting other places for me now where i cant just delete comma..)

davecoates (approved):

PR #19, opened 2020-01-15
feat: bound viewmodel fields

davecoates:
affects: @prestojs/ui, @prestojs/viewmodel

See https://gitlab.internal.alliancesoftware.com.au/alliance/presto/issues/22

There was some discussion as to whether we actually want this to be part of the API... so we can discard this if we think it's a bad idea.

----

This introduces concept of bound ViewModel fields which give you access
to a field via a record instance with the addition of the 'value' and
'isBound' properties. This makes it more concise when both the value and
field are required:

```js
<FieldFormatter field={User.fields.email} value={user.email} />
```

becomes

```js
<FieldFormatter field={user._f.email} />
```

### DIFF
--- a/django-testbed/frontend/src/views/UserListView.js
+++ b/django-testbed/frontend/src/views/UserListView.js
@@ -47,20 +47,14 @@ export default function UserListView() {
                     {records.map(record => (
                         <tr key={record._pk}>
                             <td>
-                                <FieldFormatter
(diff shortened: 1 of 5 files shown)


levic:
The implementation looks fine, but is missing high-level documentation

(This is potentially part of more high-level discussion about documentation)

PR #21, opened 2020-01-16
Feature/drf parsers and renderers

fanglong-alliance:
to be used after #13's merged

### DIFF
--- a/python-packages/presto_drf/presto_drf/parsers.py
+++ b/python-packages/presto_drf/presto_drf/parsers.py
@@ -1,6 +1,6 @@
+from distutils.util import strtobool
 import json
 
-from allianceutils.util.camel_case import camel_to_underscore
 from allianceutils.util.camel_case import underscoreize
 from django.conf import settings
 from rest_framework.exceptions import ParseError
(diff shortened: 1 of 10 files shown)


python-packages/presto_drf/presto_drf/renderers.py
```diff
@@ -0,0 +1,10 @@
+from allianceutils.util import camelize
+from rest_framework.renderers import JSONRenderer
+
+
+class CamelCaseJSONRenderer(JSONRenderer):
+    def render(self, data, *args, **kwargs):
+        print("===>", data)
+        data = camelize(data)
+        print("-->", data)
```
levic: Leftover debug code

python-packages/presto_drf/presto_drf/parsers.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    def parse(self, stream, media_type=None, parser_context=None):
+        data_and_files = super().parse(stream, media_type, parser_context)
+        request = parser_context["request"]
+        prev_mutable = data_and_files.data._mutable
+        data_and_files.data._mutable = True
+        if request.META.get(
+            "HTTP_X_RADCOMBINEDFORMPOST"
+        ):  # TODO - Confirm this's what we want to call the header? ie, RAD?
```
levic: a value of `"0"` is going to be considered true, but a value of `""` won't. That seems surprising.

python-packages/presto_drf/presto_drf/parsers.py
```diff
@@ -0,0 +1,65 @@
...
+            data = stream.read().decode(encoding)
+            return underscoreize(json.loads(data))
+        except ValueError as exc:
+            raise ParseError(f"JSON parse error") from exc
+
+
+class CamelCaseMultiPartParser(MultiPartParser):
+    """
+    If we have received any files along with other json data it gets converted
+    on frontend into multi-part post with a single key for all JSON data (formData)
```
levic: Why does this happen? Is it not possible to treat it as a vanilla multipart form?

python-packages/presto_drf/presto_drf/parsers.py
```diff
@@ -0,0 +1,65 @@
...
+            regular_form_data = json.loads(
+                data_and_files.data.get("formData"), object_pairs_hook=hook
+            )
+            return regular_form_data
+
+        for key, value in data_and_files.data.items():
+            new_key = camel_to_underscore(key)
+            if new_key != key:
+                data_and_files.data.pop(key)
+            data_and_files.data[new_key] = underscoreize(value)
```
levic: I'm not understanding something: keys make sense, but why are the values being changed?

python-packages/presto_drf/presto_drf/parsers.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class CamelCaseJSONParser(JSONParser):
+    def parse(self, stream, media_type=None, parser_context=None):
+        parser_context = parser_context or {}
+        encoding = parser_context.get("encoding", settings.DEFAULT_CHARSET)
+
+        try:
+            data = stream.read().decode(encoding)
+            return underscoreize(json.loads(data))
```
levic: One of the reasons the alliance utils `underscorize` was written was to be able to differentiate keys that are field names (and should be mangled) from keys that are mappings of a value1 => value2 (so value1 shouldn't be changed).

There doesn't seem to be any way to implement that here, even with subclassing. At minimum the underscorize() invocation should be through a method that you can override if needed

python-packages/presto_drf/presto_drf/parsers.py
davecoates: No. Normal json data can still be nested so it comes through serialised under one field

python-packages/presto_drf/presto_drf/parsers.py
davecoates: No idea... can't see any use for this section at all - should be falling back to super I believe ie. only custom thing happens when the header is set

(this came from djrad)

python-packages/presto_drf/presto_drf/parsers.py
davecoates: I'll change it to use strtobool

python-packages/presto_drf/presto_drf/parsers.py
davecoates: Done

davecoates:
@levic changes made

python-packages/presto_drf/presto_drf/tests/test_parsers.py
```diff
@@ -0,0 +1,92 @@
...
+        # Numbers in keys were problematic with old djangorestframework_camel_case.parser.CamelCaseJSONParser
+        # Test that we preserve whatever the input convetion was
+        result = parser.parse(
+            self.bytes('{"keyWithNumeric_1":{"keyWithNumeric2":1,"key":2}}')
+        )
+        self.assertEqual(
+            result, {"key_with_numeric_1": {"key_with_numeric2": 1, "key": 2}}
+        )
+
+    def test_camel_case_json_parser_underscorize_ignore(self):
```
levic: Minor since it's in a test case but this uses underscorize whereas everything else uses underscor**e**ize

python-packages/presto_drf/presto_drf/parsers.py
levic: (Assuming this was removed; looks like the history was rebase so I can't actually see the full context of this comment)

levic:
looks good to merge apart from that one typo

PR #23, opened 2020-01-17
feat: react-navi w/ breadcrumb prototype

fanglong-alliance:
affects: django-testbed, @prestojs/routing, @prestojs/ui-antd, @prestojs/ui


davecoates:
I'm going to close this now as I don't think we need it (may revisit it down the track if we come back to using react-navi or other frontend routing solution)

PR #24, opened 2020-01-21
Feature/issue23 codegen

fanglong-alliance:


python-packages/presto_codegen/presto_codegen/config.py
```diff
@@ -0,0 +1,58 @@
...
+    # These files are generated only if they do not exist and will never be overwritten by codegen
+    generate_descendant_to = None
+
+    # Additional Backend->Frontend field maps eg. {'BooField': 'FarField'};
+    #  passing existing names overwrites default settings eg. {'TextField': 'CharField'};
+    # in addition to this ^ usage, you can also pass a function as value of the map, which can be used to override entire get_field_def for field specified, eg:
+    #  {'BooField': lambda attr_name, field: {'name': attr_name, 'type': 'FarField', 'attributes': {'random': 'stuff} } }
+    # attr_name and field would be passed onto func; expected return is a dict w/ 'name', 'type' and optionally 'attributes'.
+    field_maps = {}
+
```
fanglong-alliance, line 27: Not-yet addressed comment from Gitlab MR:


is the attribute stuff necessary if we are adding methods to this class? Would it be clearer to do it as a method you override?
```
def get_attributes_for_field(self, field_name, field):
    if field_name == 'BooField': ....
```
I'm not actually sure what's a nicer interface. Feels a bit weird defining a dict of lambdas (wouldn't feel weird in JS though... )
Another thought is that this is kind of specific to the thing you are registering. Currently it will be a serializer field because we are registering a serializer but what about if we allow registering a model directly? Or a graphql node?
@levic  interested in your thoughts here

fanglong-alliance:
Also needed: test cases

(Will do once this becomes stable & have idea on what to cover)

davecoates:
Closing as we've done this in the template project now instead

PR #30, opened 2020-04-08
feature: add replaceStateControllers

davecoates:
This change is so we can maintain the identify of the paginator instance and just switch out the state as required.

This is desirable so can write hooks that use the paginator as a dependency without it changing every render

### DIFF
--- a/js-packages/@prestojs/rest/src/Paginator.ts
+++ b/js-packages/@prestojs/rest/src/Paginator.ts
@@ -17,6 +17,9 @@ export default class Paginator<State, InternalState> implements PaginatorInterfa
     setCurrentState: (set: State) => void;
     setInternalState: (set: InternalState) => void;
 
+    /**
+     * @see documentation for `replaceStateControllers` for what `currentStatePair` and `internalStatePair` are
+     */
     constructor(currentStatePair = null, internalStatePair = null) {
         if ((currentStatePair || internalStatePair) && !(currentStatePair && internalStatePair)) {
             throw new Error(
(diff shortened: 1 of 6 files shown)


js-packages/@prestojs/rest/src/Paginator.ts
```diff
@@ -17,14 +17,29 @@ export default class Paginator<State, InternalState> implements PaginatorInterfa
     setCurrentState: (set: State) => void;
     setInternalState: (set: InternalState) => void;
 
+    constructor(currentStatePair = null, internalStatePair = null) {
```
levic, line 23: There should be some documentation explaining what these are (even if it's just a `@see ...` ). In particular, is it documented somewhere that the currentStatePair can be used to specify whether to store state in a hidden way or  to store it in the URL? If so, refer back to that

js-packages/@prestojs/rest/src/Paginator.ts
```diff
@@ -17,14 +17,29 @@ export default class Paginator<State, InternalState> implements PaginatorInterfa
...
+    }
+
     /**
-     * Paginator received 2 tuples of a state and state setter pair. This is expected to
+     * Paginator receives 2 tuples of a state and state setter pair. This is expected to
      * match the same interface as `useState` in React. The following is a valid simple usage:
      *
      * ```js
-     * new Paginator(useState(), useState());
+     * const paginator = new Paginator(useState(), useState());
```
levic, line 39: See comments above re internal vs URL state. Should add a link to wherever it is documented how to use the URL for state

levic:
Looks good other than the explanations re state

levic:
@davecoates I can't assign this back to you because I'm only a Member of the prestojs organisation -- can you give me ⚡MORE POWER? ⚡

davecoates:
Header comments updated

PR #35, opened 2020-04-14
Fix/field widget props

davecoates:
Each commit is independent - bunch of random fixes

### DIFF
--- a/js-packages/@prestojs/final-form/src/FormField.tsx
+++ b/js-packages/@prestojs/final-form/src/FormField.tsx
@@ -5,7 +5,7 @@ import { Field as FinalFormField, FieldProps } from 'react-final-form';
 
 type FormFieldProps<T> =
     | FieldProps<any, any>
-    | (Omit<FieldProps<any, any>, 'name'> & { field: Field<T> });
+    | (Omit<FieldProps<any, any>, 'name'> & { field: Field<T>; widgetProps?: Record<any, any> });
 
 /**
  * Wrapper around Field from react-final-form that determines the widget to use based on the field.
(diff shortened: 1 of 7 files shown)


js-packages/@prestojs/viewmodel/src/__tests__/ViewModel.test.ts
```diff
@@ -267,17 +267,17 @@ test('should validate primary keys are provided', () => {
...
-        "Primary key(s) 'id1' was provided but was null or undefined, Missing value(s) for primary key(s) 'id2'"
+        /Primary key\(s\) 'id1' was provided but was null or undefined, Missing value\(s\) for primary key\(s\) 'id2'/
     );
     expect(() => new A({ id2: null })).toThrowError(
-        "Primary key(s) 'id2' was provided but was null or undefined, Missing value(s) for primary key(s) 'id1'"
+        /Primary key\(s\) 'id2' was provided but was null or undefined, Missing value\(s\) for primary key\(s\) 'id1'/
     );
     expect(() => new A({ id1: null, id2: null })).toThrowError(
-        "Primary key(s) 'id1', 'id2' was provided but was null or undefined"
+        /Primary key\(s\) 'id1', 'id2' was provided but was null or undefined/
```
levic: Why replace a static string with a static regex?

davecoates: string has to be full match, regex partial

I originally added some extra text to end of that error but I think I decided to roll that back

levic:
Looks fine (but am trusting that most of it, eg extra potential types is needed)

PR #36, opened 2020-04-14
fix: update ViewModel types to better support typed data

davecoates:


davecoates:
Can ignore this now replacing with factory function

PR #37, opened 2020-04-17
feature: Add ViewModelFactory

davecoates:
First part of rewrite


js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+    model: ViewModelConstructor<O>,
+    fields: O
+): ['id', NumberField] {
+    return ['id', fields.id || new NumberField()];
+}
+
+// Overloads here are so we can more accurately type the primary key and fields (specifically for default case we can
+// add the implicit 'id' field that will be created)
+export default function ViewModelFactory<O extends FieldsMapping>(
+    options: ViewModelOptionsPkFieldNameSingle<O>
```
davecoates: hmmm actually now I think about usage a bit more this should probably be split into two arguments - `fields` which is always provided and then the options.

pretty minor change so I'll address that later if you are in agreement

davecoates: actually I'm going to do it now

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+ *     lastName: new CharField(),
+ * };
+ * // Options are all optional
+ * const options = {
+ *     // If not specified will create a default field called 'id'
+ *     pkFieldName: 'userId',
+ * };
+ * class User extends ViewModel(fields, options) {
+ *     // Optional; default cache is usually sufficient
+ *     static cache = new MyCustomCache();
```
levic, line 46: How does does this interact with inheritance?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+ *
+ *     // Used to describe a single user
+ *     static label = 'User';
+ *     // User to describe an indeterminate number of users
+ *     static labelPlural = 'Users';
```
levic, line 51: Are there any defaults for label/labelPlural?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+ *     userId: new IntegerField({ label: 'User ID' })
+ *     firstName: new CharField({ label: 'First Name' }),
+ *     // label is optional; will be generated as 'Last name'
+ *     lastName: new CharField(),
+ * };
+ * // Options are all optional
+ * const options = {
+ *     // If not specified will create a default field called 'id'
+ *     pkFieldName: 'userId',
+ * };
```
levic, line 43: Does `pkFieldName` to be in `options` or can it be a static property of the `User` class like `label`?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+    new <IncomingFieldsType extends FieldMappingType = FieldMappingType>(
+        data: FieldDataMappingRaw<IncomingFieldsType>
+    ): ViewModelInterface<FieldMappingType, IncomingFieldsType, PkFieldType, PkType>;
+
+    /**
+     * The bound fields for this ViewModel. These will match the `fields` passed in to `ViewModel` with the
+     * following differences:
+     * - If a primary key is created for you this will exist here
+     * - All fields are bound to the created class. This means you can access the `ViewModel` class from the field on
+     *   the `parent` property, eg. `User.fields.email.parent === User` will be true.
```
levic, line 119: Consider `model` instead of `parent` (this is what django uses -- see `Field.contribute_to_class()`); it's not really a parent-child relationship, is more of a container/contained relationship

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+     * @return A new ViewModel class with fields modified according to `newFields`.
+     */
+    augment<P extends FieldsMappingOrNull>(
+        newFields: P
+    ): ViewModelConstructor<FieldMappingType & P, PkFieldType, PkType>;
+}
+
+type GetImplicitPkFieldCompound<O extends FieldsMapping> = (
+    model: ViewModelConstructor<O>,
+    fields: O
```
levic: Super minor style question: is `O` standard? Most of the things i've seen are `T` (for Type).

You've got a bit of a mix going on here (eg the previous class used `P`), and later on `T` is also used.

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+    extends ViewModelOptions<O> {
+    getImplicitPkField: GetImplicitPkFieldSingle<O>;
+}
+
+interface ViewModelOptionsGetImplicitPkFieldCompound<O extends FieldsMapping>
+    extends ViewModelOptions<O> {
+    getImplicitPkField: GetImplicitPkFieldCompound<O>;
+}
+
+function defineRequiredGetter(base: {}, name: string, errorMessage: string): void {
```
levic, line 221: Can you add a short 1-line comment for what this does

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+                value: label,
+            });
+        },
+    });
+}
+
+/**
+ * Generate a label for a field based on its name
+ */
+function generateFieldLabel(name: string): string {
```
levic: `generateFieldLabel` => something like `{get|generate}{Default|Implicit}FieldLabel`?

(There's no implicit labels for ViewModels is there? If so, drop the Field and have the ViewModel use the same function too since the logic should be the same)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+): ViewModelConstructor<O, string, SinglePrimaryKey>;
+export default function ViewModelFactory<O extends FieldsMapping>(
+    options: ViewModelOptionsGetImplicitPkFieldCompound<O>
+): ViewModelConstructor<O, string[], CompoundPrimaryKey>;
+export default function ViewModelFactory<O extends FieldsMapping>(
+    options: ViewModelOptions<O>
+): ViewModelConstructor<{ id: NumberField } & O, 'id', string | number>;
+export default function ViewModelFactory<O extends FieldsMapping>(
+    options: ViewModelOptions<O>
+): ViewModelConstructor<any, any> {
```
levic, line 351: Can we talk about what's going on here?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+                        ', '
+                    )}`;
+                    if (isDev()) {
+                        throw new Error(msg);
+                    } else {
+                        console.warn(msg);
+                    }
+                };
+            }
+            Object.defineProperty(this, fieldName, definition);
```
levic: How expensive are creating getters/setters?

This seems like a lot of extra work -- I would have expected they could be defined once for the class and not for each ModelView instance

(also applies to the `Object.defineProperties()` below)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+                // TODO: Should extra keys in data be a warning or ignored?
+                console.warn(
+                    `Received value for key ${key}. No such field exists on ${this._model.name}`
+                );
+            }
+        }
+
+        this._assignedFields = assignedFields;
+        this._assignedFields.sort();
+        this._data = freezeObject(assignedData);
```
levic, line 426: Can you chat to me about usage; if this is frozen then what's the correct way to overwrite value(s) in a ModelView instance?

Also for documentation: concept of partial records

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+    clone(
+        fieldNames?: string[]
+    ): ViewModelInterface<FieldMappingType, InstanceFieldMappingType, PkFieldType, PkType>;
+    /**
+     * Returns the primary key value(s) for this instance. This is a shortcut
+     * for using the primary key field name(s) directly.
+     */
+    readonly _pk: PkType;
+    readonly _data: FieldDataMapping<InstanceFieldMappingType>;
+    readonly _assignedFields: (keyof InstanceFieldMappingType)[];
```
levic, line 102: What is `_data` and `_assignedFields`? (I may be misunderstanding these, may affect some of my later comments)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+                }
+
+                // Always clone primary keys
+                const pkFieldNames = this._model.pkFieldNames;
+                pkFieldNames.forEach(name => (data[name] = this[name]));
+
+                // I don't know how to type this, error is:
+                // TS2322: Type 'ViewModel' is not assignable to type 'this'.
+                // 'ViewModel' is assignable to the constraint of type 'this', but 'this' could be instantiated with a different subtype of constraint 'ViewModel'.
+                // I could type return as `ViewModel` but then usages of it will be wrong (eg. clone() will be from a more general ViewModel to a specific implementation)
```
levic: "eg. clone() will be from a more general ViewModel to a specific implementation"

do you mean

"we want clone() to return a specific implementation type instead of a general ViewModel"?

(Was a bit unclear what you were saying)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+        ? SinglePrimaryKey
+        : typeof options.pkFieldName extends string[]
+        ? CompoundPrimaryKey
+        : typeof options.getImplicitPkField extends GetImplicitPkFieldCompound<O>
+        ? CompoundPrimaryKey
+        : typeof options.getImplicitPkField extends GetImplicitPkFieldSingle<O>
+        ? SinglePrimaryKey
+        : SinglePrimaryKey;
+
+    // This is the constructor function for the created class
```
levic: Why is this not just a regular `constructor()` on a regular class?

ie:

```
function ViewModelFactory() {

   class GenericViewModel {
      constructor() {
          ...
      }
   }

  GenericViewModel.name = ...;

  return GenericViewModel;
}
```

levic (changes requested):

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+}
+
+function defaultGetImplicitPkField<O extends FieldsMapping>(
+    model: ViewModelConstructor<O>,
+    fields: O
+): ['id', NumberField] {
+    return ['id', fields.id || new NumberField()];
+}
+
+// Overloads here are so we can more accurately type the primary key and fields (specifically for default case we can
```
levic: "Overloads here" => "Using very strongly typed overloads instead a single type with optional properties"

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,591 @@
...
+
+        this._assignedFields = assignedFields;
+        this._assignedFields.sort();
+        this._data = freezeObject(assignedData);
+
+        // Add getters for fields along with some traps for bad usage
+        // - If attempts to set a field throw an error in dev or log warning otherwise
+        // - If attempts to get a field that does exist but wasn't passed in `data` then
+        //   throw an error in dev or log a warning otherwise.
+        for (const fieldName of Object.keys(fields)) {
```
levic: Can you double check whether this needs to be on each instance instantiation or whether the definition can go once on the class

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: No not possible unless we require a `name` to be passed in for the created class itself

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: As discussed must be passed in as we create / validate fields based on this

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: hehe I went through so many iterations of type definitions that it's mainly a side effect of that + examples I copied and pasted

will clean it up

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: ViewModel instances are immutable - you don't overwrite any values on it ever (this sounds familiar to discussions we had on original `ViewModel` implementation...)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
levic: A thought bubble: is it feasible to define field properties as regular JS object properties and use typescript's `readonly` to enforce immutability?

- regular JS, esp 3rd party libs (eg lodash) would still see properties as writable
- is it even possible in typescript to inject read-only properties into your new class based on the incoming field names?

davecoates: > A thought bubble: is it feasible to define field properties as regular JS object properties and use typescript's `readonly` to enforce immutability?
> 
> * regular JS, esp 3rd party libs (eg lodash) would still see properties as writable
> * is it even possible in typescript to inject read-only properties into your new class based on the incoming field names?

`readonly` doesn't change anything at runtime (if that's what you meant?). any properties added with `defineProperty` are by default immutable.

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: `_data` is the data for this object. the getters for each field read from this. i'll add a comment indicating you usually don't need to access this directly.

`_assignedFields` is the fields set on this object (eg. partial record with these fields set)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: I've moved this to the prototype

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: I've removed this entirely - seemed to be remnant of original class implementation and error no longer exists

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: couldn't get type definitions to work. i think particularly the indexing signature

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: workrs - test cases expanded to more comprehensively test this

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: Ok this isn't something introduced in this branch so i've raised #40 to rename it

django-testbed/frontend/src/models/generated/BaseUser.js
```diff
@@ -1,23 +1,21 @@
 import { CharField, EmailField, ViewModel, NumberField } from '@prestojs/viewmodel';
 
-export default class Base extends ViewModel {
+export default class Base extends ViewModel({
+    id: new NumberField({
+        label: 'Id',
+    }),
+    // eslint-disable-next-line @typescript-eslint/camelcase
```
levic, line 7: Why the python naming convention here? I thought we would have done camel case conversion in the code gen

js-packages/@prestojs/codegen/tsconfig.lint.json
```diff
@@ -4,7 +4,8 @@
   "compilerOptions": {
     "composite": true,
     "rootDir": "./src",
-    "tsBuildInfoFile": "build/tsconfig.tsbuildinfo"
+    "tsBuildInfoFile": "/tmp/tsc-output/@prestojs/codegen.tsbuildinfo",
+    "outDir": "/tmp/tsc-output/@prestojs/codegen"
```
levic, line 8: Should that hardcoded `/tmp` be there? (appears in other places too)

tsconfig.base.json
```diff
@@ -10,12 +10,19 @@
     "baseUrl": ".",
     "paths": {
       "@prestojs/viewmodel": ["./js-packages/@prestojs/viewmodel/src/"],
+      "@prestojs/viewmodel/*": ["./js-packages/@prestojs/viewmodel/src/*"],
```
levic, line 13: Is that right? You have to specify both `path` and `path/*`?

js-packages/@prestojs/viewmodel/src/__tests__/ViewModelCache.test.ts
```diff
@@ -34,23 +29,19 @@ test('should cache records from record instance', () => {
...
+        id: F('id'),
+    }) {}
 
     const record1Data = { id: 5 };
 
     const record1 = Test1.cache.add(record1Data);
 
     expect(record1).toBeInstanceOf(Test1);
-    expect(record1 instanceof ViewModel ? record1.toJS() : false).toEqual(record1Data);
+    expect(isViewModelInstance(record1) ? record1.toJS() : false).toEqual(record1Data);
```
levic, line 44: This seems like a step backwards in intuitiveness/discoverability; I assume this is because it's a factory generating the base class? Is it possible to preserve the original behaviour?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,699 @@
...
+function defaultGetImplicitPkField<T extends FieldsMapping>(
+    model: ViewModelConstructor<T>,
+    fields: T
+): ['id', NumberField] {
+    return ['id', fields.id || new NumberField()];
+}
+
+const IS_VIEW_MODEL = Symbol.for('@prestojs/IS_VIEW_MODEL');
+
+export function isViewModelInstance(view: any): view is ViewModelInterface<any, any> {
```
levic, line 309: See comment elsewhere; is it possible to avoid having this function at all so that a vanilla `instanceof` works?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,699 @@
...
+    return !!(view && view.constructor && view.constructor[IS_VIEW_MODEL]);
+}
+
+export function isViewModelClass(view: any): view is ViewModelConstructor<any> {
```
levic, line 313: As above re `instanceof` (although less important in this case -- I would regard looking at a viewmodel class (as opposed to instance) as a form of meta magic introspection and wouldn't necessarily expect that it would work, particularly since the base class comes from a factory function)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
levic: Can you add a comment to this effect explaining why this is not a regular ES6 `class`

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -0,0 +1,699 @@
...
+    });
+    Object.defineProperties(_Base.prototype, properties);
+
+    // Store bound fields and primary key name for all models in the hierarchy
+    const boundFields: Map<
+        ViewModelConstructor<FinalFields>,
+        [FinalFields, string | string[]]
+    > = new Map();
+
+    function _bindFields(modelClass: ViewModelConstructor<T>): [FinalFields, string | string[]] {
```
levic, line 544: Please add a function header for this (For devs who are debugging trying to figure out what's going on). Not so much interesting in arguments or return values but a higher-level explanation of instance field binding

levic: (or maybe @see boundFields and put the explanation there)

levic: Can you clarify that this is lazily evaluated? (seems to be only used in getters)

Am thinking about the use case of getting eg 10,000 results in an ajax request but only displaying the first 100: would be good to defer the evaluation of this until you actually request a field value

levic (changes requested):

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
levic: Yeah, I was thinking about just letting TS enforce the readonly rather than the JS runtime.

Have thought a bit more about it and is probably going to be too dangerous as the consuming applications may not be TS and someone will end up modifying data directly and then complain about bugs in presto

django-testbed/frontend/src/models/generated/BaseUser.js
davecoates: because we don't have the camel case conversion on backend yet

js-packages/@prestojs/codegen/tsconfig.lint.json
davecoates: for the linting the files have to be emitted somewhere otherwise has to do full compilation everytime (slow) so workaround is to output them somewhere - so currently hardcoded to /tmp

tsconfig.base.json
davecoates: yeah `import { ... } from '@prestojs/viewmodel'` doesn't work without it and doesn't support regex to make slash optional

js-packages/@prestojs/viewmodel/src/__tests__/ViewModelCache.test.ts
davecoates: The only possibility would be to make the factory class itself the base... maybe.

I would personally find that a bit confusing knowing the the factory function itself also serves as the base class but maybe...? what do you think?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: if we can change `instanceof` for instance to work we can probably change this (eg. `B instanceof A.constructor`)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: yes it's only evaluated when `fields` or `pkFieldName` are accessed

but remember this is on the class, not the instance - so unless those 10,000 results are 10,000 different viewmodels it doesn't matter

js-packages/@prestojs/viewmodel/src/__tests__/ViewModelCache.test.ts
levic: Yeah, if that makes `instanceof` work it would be worth it.

I expect that someone looking at the library code would probably think "wtf is going on here" but as long as you add a `// comment` around it explaining why we do it that way then that's a net positive

levic: To clarify, my understanding is that you would end up with something that looked like:

- `ViewModelInterface` -- typescript definition
- `ViewModel` -- an ES6 base class (may be a placeholder that doesn't really do much)
- `ViewModelFactory` -- a class generator that returns a class that is a descendant of ViewModel

(I say "looks like" because ViewModel may or may not be declared as an actual ES6 class, it just needs to look like one so that `instanceof` works)

PR #43, opened 2020-05-01
Fix/misc fixes may5 2020

davecoates:


davecoates:
@levic review commit by commit, last commit is just prettier2

js-packages/@prestojs/ui/src/UiProvider.tsx
```diff
@@ -19,9 +19,15 @@ type GetFormatterForFieldWithNull = <FieldValue, T extends HTMLElement>(
...
 export interface FormItemProps {
+    children: React.ReactNode;
     required: boolean;
     help?: React.ReactNode;
     label?: React.ReactNode;
+    // Field name, if known. This is only applicable when a field has been passed through
+    // to a FormItem.
+    fieldName?: string;
+    // Any other props
+    [x: string]: any;
```
levic, line 30: Why is this necessary if a typescript interface is already only a minimum set of fields that must be present? Just for clarity?

js-packages/@prestojs/rest/src/usePaginator.ts
```diff
@@ -13,9 +13,26 @@ import { PaginatorInterface, PaginatorInterfaceClass } from './Paginator';
...
+): T;
+export default function usePaginator<T extends PaginatorInterface, PaginatorState>(
+    paginatorClassOrEndpoint: null,
+    currentStatePair?: [PaginatorState | undefined, (value: PaginatorState) => void]
+): null;
+// This is only necessary because the type is inferred as PaginatorInterfaceClass<T> | null here
+// instead of refining it based on endpoint
+// export function ExampleComponent({ endpoint }: { endpoint: Endpoint | null }) {
+//     usePaginator(endpoint && endpoint.getPaginatorClass());
+// }
```
levic, line 27: Does this need a test case or is it simple enough that it's not worth it?

js-packages/@prestojs/viewmodel/src/fields/DateTimeField.ts
```diff
@@ -8,11 +8,20 @@ import Field from './Field';
...
         return new Date(value);
     }
+    normalize(value: string | Date): Date | null {
+        if (value instanceof Date) {
+            return value;
+        }
+        if (Number.isNaN(Date.parse(value))) {
+            return null;
+        }
+        return new Date(value);
```
levic, line 25: This data should always be coming from an ajax call so should be ISO format?

We're not going to run into cases where the user's browser locale changes how it's interpreted?

js-packages/@prestojs/ui/src/UiProvider.tsx
davecoates: It's necessary - interface is not a minimum set of fields, eg: https://www.typescriptlang.org/play/?ssl=14&ssc=38&pln=9&pc=1#code/JYOwLgpgTgZghgYwgAgILIN4FgBQz-IhwC2EAXMgM5hSgDmA3LgL664ID2I1ycF6AXky4ChEuWQBySNUkAaEQQgAPGnADSEAJ4VJcSgBMY8lriA

js-packages/@prestojs/viewmodel/src/fields/DateTimeField.ts
davecoates: yes to the first part

I have no idea about the second question... 🤔

levic: I was thinking about the user typing a date into a freeform text field -- the date widget should take care of that normalisation shouldn't it?

js-packages/@prestojs/rest/src/usePaginator.ts
davecoates: test added

js-packages/@prestojs/viewmodel/src/fields/DateTimeField.ts
davecoates: See #45. I'm going to merge this as is and we can revisit (it's already in use like this elsewhere, eg in DateTime.ts).

levic (approved):

PR #52, opened 2020-05-13
Feature/use view model cache extra args

davecoates:
- Adds ability to pass extra args through to selector
- Adds ability to change comparison function as optimisation

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/useViewModelCache.ts
+++ b/js-packages/@prestojs/viewmodel/src/useViewModelCache.ts
@@ -26,6 +26,10 @@ function shallowEqual(a?: any[], b?: any[]): boolean {
     return true;
 }
 
+function defaultEquality<T>(currentValue: T, previousValue?: T): boolean {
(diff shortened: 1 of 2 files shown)


js-packages/@prestojs/viewmodel/src/useViewModelCache.ts
```diff
@@ -1,12 +1,35 @@
...
+    ...args: U
 ) => ResultType;
 
+function shallowEqual(a?: any[], b?: any[]): boolean {
+    if (a === b) {
+        return true;
+    }
+    if (!a || !b) {
+        return false;
+    }
```
levic, line 17: ~If `a == []` and `b == []` this will return false~

Edit:JS, not python.

levic:
This looks good to me

js-packages/@prestojs/viewmodel/src/useViewModelCache.ts
davecoates: heh

PR #56, opened 2020-05-27
Feature/async choices

davecoates:
This contains changes required for async choices but doesn't have the actual changes to `Field` or widget - I'll do them as a separate branch

Note that I ported `useAsync` from djrad - you can probably skip / skim those commits.

Probably easiest to review this commit by commit

Here's an example usage for a widget based on Endpoint

```
       const listEndpoint = new PaginatedEndpoint(namedUrls.get('users-list'));
       const retrieveEndpoint = new Endpoint(namedUrls.get('users-detail'));
       const paginator = usePaginator(listEndpoint);
        const result = useAsyncLookup({
            query,
            accumulatePages,
            execute: async props => (await listEndpoint.execute(props)).result,
            paginator,
        });
       // This is the currently selected value which may be in result returned above or
       // may need to be fetched
       const currentValueResolved = useAsyncValue({
            existingValues: result.result,
            id: currentValue || null,
            resolve: async id => retrieveEndpoint.execute({ urlArgs: { id } }),
        });
```

To discuss:

1) Naming of `useAsyncValue` and `useAsyncLookup` - don't think these names are great so expect them to change. Any suggestions?

2) Currently pagination and hence `useAsyncLookup` is in `@prestojs/rest` but don't really need to be - it's more generic than that. Where should we move these to? I think the realistic options are either `@prestojs/util` or a new module.


js-packages/@prestojs/rest/src/CursorPaginator.ts
```diff
@@ -190,4 +190,11 @@ export default class CursorPaginator extends Paginator<
             this.setCurrentState({ ...this.currentState, pageSize });
         }
     }
+
+    /**
+     * Returns true if there's more results after the current page
+     */
+    hasNextPage(): boolean {
```
levic, line 197: Is this always going to be possible to know?

What if eg you're on the second page of 20 and there are only 40 results in total? Depending on the implementation, wouldn't this make you potentially have to fetch 2 pages from the database in order to know that it's possible to fetch the next page?

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -35,6 +35,21 @@ type EndpointOptions = ExecuteInitOptions & {
...
+     * A function to resolve the URL. It is passed the URL pattern object, any
+     * arguments for the URL and any query string parameters.
+     *
+     * If not provided defaults to:
+     *
+     * ```js
+     * urlPattern.resolve(urlArgs, { query });
+     * ```
+     */
+    resolveUrl?: (
```
levic, line 48: Why do you provide the ability to override this? It doesn't seem to be used

js-packages/@prestojs/rest/src/InferredPaginator.ts
```diff
@@ -47,6 +47,9 @@ export type PaginatorState =
  */
 export default class InferredPaginator implements PaginatorInterface {
     __paginator?: CursorPaginator | PageNumberPaginator | LimitOffsetPaginator;
+    get responseSet(): boolean {
```
levic: This name is confusing: I thought it was a set of responses and went looking to figure out what on earth it was.

Alternatives: `responseIsSet()`, `isResponseSet()`, `isResponsePresent()`, `responsePresent()` etc

(also applies in other places)

js-packages/@prestojs/rest/src/index.ts
```diff
@@ -7,6 +7,7 @@ export { default as LimitOffsetPaginator } from './LimitOffsetPaginator';
 export { default as CursorPaginator } from './CursorPaginator';
 export { default as Paginator } from './Paginator';
 export { default as getPaginationState } from './getPaginationState';
+export { default as EndpointAsyncChoices } from './EndpointAsyncChoices';
```
levic: Where's that coming from? It doesn't seem to be in this commit.

(Is it supposed to be in a subsequent commit? Or was a forgotten export from a previous commit?)

js-packages/@prestojs/rest/src/useAsyncLookup.ts
```diff
@@ -0,0 +1,208 @@
...
+import { useEffect, useRef } from 'react';
+import { PaginatorInterface } from './Paginator';
+
+/**
+ * @expand-properties
+ */
+type UseAsyncLookupProps<T> = {
+    /**
+     * When to trigger the fetch. Defaults to `SHALLOW` which means whenever `query`
+     * or `paginator` state changes it will refetch.
```
levic: "whenever `query` or `paginator` state changes it will refetch." presumably means "whenever a shallow `query` or `paginator` comparison detects a change it will refetch."?

js-packages/@prestojs/rest/src/useAsyncLookup.ts
```diff
@@ -0,0 +1,208 @@
...
+     * Whether to accumulate pages as more results are fetched. For example
+     * if the first page of results is returned, then the next page is fetched
+     * then the combined results for the first two pages will be resolved.
+     *
+     * This resets whenever `query` changes or if pagination state changes
+     * to anything other than the next page.
+     *
+     * If this is true you must specify `paginator`.
+     */
+    accumulatePages?: boolean;
```
levic, line 35: This is a nice touch to build this in here rather than expecting the widget to do it

js-packages/@prestojs/rest/src/useAsyncLookup.ts
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * Optional paginator if response is paginated. This will be monitored for
+     * any state changes which will trigger a call to `execute`.
+     *
+     * Required if `accumulatePages` is true.
```
levic, line 40: I would be incline to actually remove that comment. It's already specified under `accumulatePages` and is just a diversion here

levic: I can't find any explicit check at runtime for where you specify `accumulatePages` but not `paginator`-- I expect that is going to be easier to diagnose than "Cannot read property 'paginator' of null"

js-packages/@prestojs/rest/src/useAsyncLookup.ts
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * True while `execute` call is in progress.
+     */
+    isLoading: boolean;
+    /**
+     * The same `paginator` passed in to `useAsyncLookup`
+     */
+    paginator: null | PaginatorInterface;
+    /**
+     * Set to the rejected value of the promise. Only one of `error` and `response` can be set. If
```
levic: Is it worth getting typescript to enforce the mutual exclusivity of error & response? (I expect transitioning from (error=null,response=not null) to (error=not null,response=null) is painful?)

js-packages/@prestojs/rest/src/useAsyncLookup.ts
```diff
@@ -0,0 +1,208 @@
...
+     * A function to manually trigger the action. If `options.trigger` is `MANUAL`
+     * calling this function is the only way to trigger the action.
+     *
+     * This function will return a promise that resolves/rejects to same value
+     * returned by `execute`. If `accumulatePages` is set the value returned is
+     * the accumulated value.
+     */
+    run: (...args) => Promise<any>;
+    /**
+     * When called will unset result and error to null. Will not immediately trigger
```
levic: "unset result and error to null" -- why not just "set both result and error to null"?

js-packages/@prestojs/rest/src/useAsyncLookup.ts
```diff
@@ -0,0 +1,208 @@
...
+}: UseAsyncLookupProps<T>): UseAsyncLookupReturn<T> {
+    const initialRun = useRef(true);
+    const paginationState = paginator?.responseSet && paginator?.currentState;
+    const nextPaginationStateRef = useRef<{} | null>(null);
+    const lastPaginationStateRef = useRef<{} | null>(null);
+    const lastQuery = useRef(query);
+    const executeRef = useRef(execute);
+    const queryChanged = !isEqual(lastQuery.current, query);
+
+    executeRef.current = execute;
```
levic: What's this used for? It doesn't seem to ever be touched again (should the execute below really be executeRef?)

js-packages/@prestojs/rest/src/useAsyncLookup.ts
```diff
@@ -0,0 +1,208 @@
...
+    execute,
+    paginator = null,
+}: UseAsyncLookupProps<T>): UseAsyncLookupReturn<T> {
+    const initialRun = useRef(true);
+    const paginationState = paginator?.responseSet && paginator?.currentState;
+    const nextPaginationStateRef = useRef<{} | null>(null);
+    const lastPaginationStateRef = useRef<{} | null>(null);
+    const lastQuery = useRef(query);
+    const executeRef = useRef(execute);
+    const queryChanged = !isEqual(lastQuery.current, query);
```
levic, line 139: Can you chat to me about the useRef()s here -- are they required for the same reason as the comments for shouldResetAccumulatedValues below? (If so, can you hoist that comment to the top)

js-packages/@prestojs/rest/src/useAsyncLookup.ts
```diff
@@ -0,0 +1,208 @@
...
+ * If the response is paginated you can pass `paginator`. Whenever the paginator state
+ * is changed the function will be called unless `trigger` is `MANUAL`. You can pass
+ * `accumulatePages` to accumulate results for sequential pages returned from `execute`.
+ * This is useful to implement things like infinite scroll. If a non-sequential page
+ * is accessed or `query` changes then accumulated results will be cleared.
+ *
+ * @extract-docs
+ */
+export default function useAsyncLookup<T>({
+    trigger,
```
levic: Shouldn't this default to 'SHALLOW'?

It's kind of implicit because all the checks are against 'MANUAL', so undefined is treated the same as SHALLOW but this is not clear unless you look at the code

js-packages/@prestojs/rest/src/useAsyncValue.ts
```diff
@@ -0,0 +1,249 @@
...
+     */
+    getId?: (item: T) => U;
+    /**
+     * When to trigger the fetch. Defaults to `DEEP` which means whenever `id`
+     * or `ids` changed it will refetch if the value hasn't already been resolved.
+     *
+     * If set to `MANUAL` nothing will happen until it changes to `DEEP`. You
+     * can use this to defer execution until the value is required.
+     */
+    trigger?: 'MANUAL' | 'DEEP';
```
levic, line 45: Why DEEP? If all you're doing is comparing a scalar or a list of scalars isn't that shallow? (seems strange that useAsyncLookup is MANUAL vs SHALLOW whereas this is MANUAL vs DEEP)

js-packages/@prestojs/rest/src/useAsyncValue.ts
```diff
@@ -0,0 +1,249 @@
...
+export default function useAsyncValue<T, U extends Id>(
+    props: CommonProps<T, U> & {
+        id?: U | null;
+        ids?: U[] | null;
+        resolve: (idOrIds: U | U[]) => Promise<T[] | T>;
+    }
+): UseAsyncValueReturn<T | T[]> {
+    const { id, ids, getId, resolve, trigger } = props;
+    if (id && ids) {
+        throw new Error("Only one of 'id' and 'ids' should be provided");
```
levic, line 167: Can't TS enforce this? (or is this check for non-TS clients?)

js-packages/@prestojs/rest/src/useAsyncValue.ts
```diff
@@ -0,0 +1,249 @@
...
+    }
+    const idRef = useRef(id || ids);
```
levic, line 170: Same comments above re useRef

esp what's the point of `resolveRef` if you just assign it to `resolve` on every render?

js-packages/@prestojs/rest/src/useAsyncValue.ts
```diff
@@ -0,0 +1,249 @@
...
+                const hashedId = hashId(id);
+                const cachedItem = cache.current.get(hashedId);
+                if (cachedItem) {
+                    valueInChoices = cachedItem;
+                }
+            }
+        }
+    }
+    const isValueMissing = (id || ids) && !valueInChoices;
+    const execute = useCallback(id => resolveRef.current(id), []);
```
levic: If `resolve` changes then won't `useCallback` not invalidate?

(Similarly, what invalidates `cache`?)

js-packages/@prestojs/util/src/identifiable.ts
```diff
@@ -0,0 +1,124 @@
...
+      }
+    | {
+          id?: Id;
+          _pk: Id;
+      };
+
+/**
+ * Check if a value conforms to Identifiable
+ */
+export function isIdentifiable(item: any): item is Identifiable {
```
levic, line 25: should this be a template function instead of `item: any`? (also applies to `getId`)

js-packages/@prestojs/util/src/identifiable.ts
```diff
@@ -0,0 +1,124 @@
...
+        for (const name of f) {
+            const value = id[name];
+            if (
+                name.includes(HASH_SEP) ||
+                name.includes(FIELD_SEP) ||
+                (typeof value == 'string' &&
+                    (value.includes(HASH_SEP) || value.includes(FIELD_SEP)))
+            ) {
+                throw new Error(
+                    `Compound id field name or value cannot include ${HASH_SEP} or ${FIELD_SEP}`
```
levic: Rather than throwing an error can we not just add an escape mechanism? Am not worried so much about the field name, but the field value could contain binary data that will appear to work right up until it mysteriously doesn't

js-packages/@prestojs/util/src/useAsync.ts
```diff
@@ -0,0 +1,439 @@
...
+ * For data fetching you usually want _automatic_ mode as you retrieve some data
+ * initially and then refetch it when some arguments change (eg. the id for a single
+ * record or the filters for a list).
+ *
+ * ## Examples
+ *
+ * Fetch and render a specified github profile
+ *
+ * ```js
+ * function getGithubUser(user) {
```
levic, line 194: I know it's redundant, but given that this is the example maybe add something along the lines

```
// we don't define this inside FollowerCount() because that will create a new function on
// every render, causing useAsync() to re-run and triggering an infinite render loop
```

js-packages/@prestojs/util/src/useAsync.ts
```diff
@@ -0,0 +1,439 @@
...
+     *
+     * **NOTE**: If changing from MANUAL then the function will be called immediately regardless
+     *
+     * **useAsync.MANUAL (default)** - only called when you explicitly call `run`
+     *
+     * **useAsync.SHALLOW** - called whenever a shallow equality check fails. Compares previous async function,
+     * and `option.args`. Passing an inline function (eg. `useAsync(() => ...)`) or an inline object
+     * to args (eg. `useAsync(fn, { args: { filter: 1 } })`) with this option will result in an
+     * infinite loop because each render dynamically creates a new object and only object identity is checked;
+     * use `useMemo` or `useCallback` in these cases.
```
levic, line 91: You never found a way to automatically detect this?

Even if it was just X runs within Y seconds then log to the console, it might help devs (not sure how you could do that without too much overhead -- maybe every 1,000 calls check the elapsed time since the last 1,000 calls started?)

levic (changes requested):

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: In our current implementations it's possible to know - LimitOffset and PageNumber both know the total number of records & Cursor knows the next cursor

js-packages/@prestojs/util/src/useMemoOne.ts
```diff
@@ -0,0 +1,80 @@
...
+type Cache<T> = {
+    inputs?: any[];
+    result: T;
+};
+
+export type ComparisonFn = (newInputs: any[], lastInputs: any[]) => boolean;
+
+/**
+ * Based on https://github.com/alexreardon/use-memo-one but supports custom
+ * comparison operator.
```
levic: Can you add a basic description that doesn't require you to look up that site,

"useMemo() but with a guarantee that results will be cached for the lifetime of the component"

Also, maybe a reference to this issue which seems to do what you want: https://github.com/alexreardon/use-memo-one/issues/11

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: It's used in the next PR:

```js
new PaginatedEndpoint(namedUrls.get('users-list'), {
        transformResponseBody: transformAndCacheUser,
        resolveUrl(urlPattern, urlArgs, query) {
            return urlPattern.resolve(urlArgs, {
                // default to pageNumber pagination if not specified
                query: { ...query, paginationType: query.paginationType || 'pageNumber' },
            });
        },
    })
```

js-packages/@prestojs/rest/src/index.ts
davecoates: It was part of original implementation that is removed in next PR

js-packages/@prestojs/util/src/useAsync.ts
```diff
@@ -0,0 +1,439 @@
...
+            `The 'args' option must be an array of arguments to pass to the async function. If you are trying to pass a single argument wrap it in an array, eg. '[<arg>]'. Received type: ${typeof args}`
+        );
+    }
+    if (![SHALLOW, MANUAL, DEEP].includes(trigger)) {
+        throw new Error(`Invalid trigger ${trigger}`);
+    }
+    // =========================================================================
+
+    const comparisonFn = comparisonByTrigger[trigger];
+    const memoizedArgs = useMemoOne(() => args, args, comparisonFn);
```
levic, line 274: Is useMemoOne really necessary? If some results are re-fetched then does that matter? (Should this behaviour be an optional param?)

js-packages/@prestojs/util/src/useAsync.ts
```diff
@@ -0,0 +1,439 @@
...
+    // We cache the lastValues so we can read then when run() is called. This allows
```
levic: then=>them

js-packages/@prestojs/util/src/useAsync.ts
```diff
@@ -0,0 +1,439 @@
...
+
+    // abortRef is used by execute to store an abort function so that if the
+    // component unmounts before the async function finishes running then it
+    // can avoid calling success / error functions that would set state.
+    const abortRef = useRef<(() => void) | null>();
+
+    // =========================================================================
+    // Effect to trigger the execute method when:
+    // 1) Parameters change
+    // 2) The actual function changes (?)
```
levic: What's the `(?)` for?

js-packages/@prestojs/rest/src/useAsyncValue.ts
davecoates: Nothing invalidates this because it has empty deps. It reads the ref at execution time so that `execute` never changes

`cache` is a ref - it will be freed when the component unmounts

js-packages/@prestojs/util/src/identifiable.ts
davecoates: Keep in mind this is only for ids... seems very unlikely to ever have binary data there?

Is it worth the effort?

js-packages/@prestojs/util/src/useAsync.ts
davecoates: Not sure if we should do this as part of lib... I think probably just add https://github.com/kentcdodds/stop-runaway-react-effects to project to detect it generally

js-packages/@prestojs/rest/src/useAsyncLookup.ts
davecoates: Hmm actually this should be `DEEP` as that's actually what the implementation checks it's just that query or paginator state are always 1 level deep (currently). I'm only using this to be consistent with `useAsync` - it's really just either MANUAL or AUTO.

I'll change to use `DEEP` instead

js-packages/@prestojs/rest/src/useAsyncLookup.ts
davecoates: ah yep will add

js-packages/@prestojs/util/src/useAsync.ts
```diff
@@ -0,0 +1,439 @@
...
+    // is called with.
+    const run = useCallback(
+        (...executeArgs) => {
+            if (
+                executeArgs.length === 1 &&
+                executeArgs[0] &&
+                typeof executeArgs[0] === 'object' &&
+                executeArgs[0].actionArgs
+            ) {
+                // TODO: This check can be removed once True Value & Southern Cross have upgraded
```
levic: Are they using this codebase or is that a cut & paste leftover?

js-packages/@prestojs/util/src/useAsync.ts
```diff
@@ -0,0 +1,439 @@
...
+                abortRef.current();
+            }
+        };
+    }, []);
+    // =========================================================================
+
+    // =========================================================================
+    // Create callback that is returned and can be called by the consumer. This
+    // just calls `execute` but optionally supports changing the arguments it
+    // is called with.
```
levic: " optionally supports changing the arguments it is called with." 

Can you explain what you mean by that?

levic (changes requested):

js-packages/@prestojs/rest/src/useAsyncLookup.ts
davecoates: sorry it's leftover from original implementation before I refactored to use `useAsync`

js-packages/@prestojs/rest/src/useAsyncLookup.ts
davecoates: executeRef needs to be removed (was part of original implementation)

`useRef` for query is just to give as a way to compare previous to current. eg. each render we compare `lastQuery.current` to `query` and then whenever we make the call to `execute` we update `lastQuery.current = query`.

js-packages/@prestojs/rest/src/useAsyncLookup.ts
davecoates: I think I can change this... I originally didn't as I needed a way to differentiate between user choosing trigger and the default... but I think I changed the expected behaviour such that the difference no long matters

js-packages/@prestojs/rest/src/useAsyncValue.ts
davecoates: Both need to be `DEEP` (see other comment). This has to be deep otherwise compound keys won't work

js-packages/@prestojs/rest/src/useAsyncValue.ts
davecoates: yeah typescript will pick this up because of overloads with `UseAsyncValuePropsSingle` and `UseAsyncValuePropsMulti` - just for runtime when not using TS

js-packages/@prestojs/rest/src/useAsyncValue.ts
davecoates: the reason is so you can read the value of the ref in the memoized `execute` function below. eg. whenever `execute` is called it uses the last value set. Using this we can support passing inline functions to `useAsyncValue`... without it you'd be forced to memoize it yourself all the time. The downside of it is if the _only_ thing that changes is the function then it won't trigger a change but for the use case here I think this is acceptable (ie. the only thing that will recalculate the value is a change in `id`/`ids` or a call to `run` directly. This is similar to `useSelector` in redux - it uses the same approach from memory.

Perhaps worth some more thought / discussion... in particular I do that here but I don't do it in `useAsync`. Perhaps `useAsync` should do it too but seems riskier to me given it's a more general purpose hook

js-packages/@prestojs/util/src/useAsync.ts
davecoates: it's necessary for the `comparisonFn` support - without it it would change every time

js-packages/@prestojs/util/src/useAsync.ts
davecoates: heh no idea too long ago... copied from djrad

js-packages/@prestojs/util/src/useAsync.ts
davecoates: copy paste

js-packages/@prestojs/util/src/useAsync.ts
davecoates: you can pass arguments to `run` when you call it to override the defaults

js-packages/@prestojs/rest/src/useAsyncLookup.ts
davecoates: hmm it's easy enough to do this the only downside is it makes the autogenerated docs harder to read, eg. they would become here:

image


vs 


image


I don't know... what do you think?

js-packages/@prestojs/rest/src/useAsyncLookup.ts
davecoates: yep better 👍

js-packages/@prestojs/util/src/identifiable.ts
davecoates: honestly not entirely sure but I don't think so, see https://www.typescriptlang.org/play/?#code/C4TwDgpgBAyglgOwOYBsIEkAmUC8UDOwAToklAD5QICuAtgEYREDcAsAFCiRQDCA9rTB9qCTFlxQA3lADaAMzgQUmAHIBDWhABcBYqQC6O+MjTiAvm3YQAHkKLAoXaOLzHUGbJX6DhorJY4AegAqYI4oYKh0BGAmOTUAY2g5PiJHcAh8RwALNQcAd2gEtQQoNWpgATy4YpQUECgbYkSHNSgROABHaghwyLhMCBi4BSYoOSIBADo+vqi5MqgAAwB9MABrJagwSch7Bps4QiyACggkKfHJ2lkANUV8gFk+QZQAMRbUkH0TzD4ErT3CBPF5KD4JSpEEAASmhc2AuQKcDqUEY7XwEGwfARTHyR2gJWWAy2Oz4e1AUDxKLR1AxmBm7AigQ4NjsDicUUGwwUanoaFw4SgQsokkFQvF4oGOn8Yol4rW6wA-NLMJY5VAzLKRbL1QNlZy1er5RsVYaJRYOEFQnMeNkIAl1lARosAG5qFA9KAJPgIFJEWhZSqcobAEZwXloPrMqy2VIOOQiCFwH1O-BYENhiMQAA8ABUAHwnOCxWg6XPQnTFiA3I7B7nhvnQUWMyULE4AQirN3IlCcfAWXag7ZweAA5Hx6AAre3AUfQqQ68VECDAahEUrxFAYs0a2XL1frp0lqYDId4GgontH6tTBVnqjUOqWMxAA for what happens if make it generic

js-packages/@prestojs/util/src/useAsync.ts
davecoates: good idea 👍

js-packages/@prestojs/rest/src/CursorPaginator.ts
levic: What were you using for the cursor implementation?

(comparison: for a python iterator you can't know whether there are more results to fetch without actually trying to do the fetch and then getting a StopIteration exception)

js-packages/@prestojs/util/src/identifiable.ts
levic: It should only be 1 line shouldn't it? (a str replace to double up any occurrences of the separator)

js-packages/@prestojs/util/src/useAsync.ts
levic: That library looks useful -- is maybe worth mentioning in the comment?

js-packages/@prestojs/util/src/identifiable.ts
levic: Am not sure what I'm seeing here -- the output looks identical for both.

I realised that I missed the `: item is Identifiable` though so TS already has all the information it needs to do subsequent checks -- this is what I was thinking would allow this function to be potentially optimised away.

davecoates: Sorry I meant the errors you get if you change it to a generic

js-packages/@prestojs/rest/src/useAsyncLookup.ts
levic: Sounds good

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: DRF 

eg. last page looks like:

image

js-packages/@prestojs/rest/src/useAsyncValue.ts
levic: I’ve wasted hours on before on react components that sometimes don’t react to props changes as you would expect.

Do you have an example of what the dev/code cost would be if we simply respond to the execute function changing without trying to add any conveniences?

davecoates: Sure, if we make it a dep then someone writes this:

```ts
const retrieveDeps = asyncChoices.useRetrieveDeps({});
const selected = useAsyncValue({
     ids: value,
     resolve: props => asyncChoices.retrieve(props, retrieveDeps),
});
```

they get runaway effect/infinite loop

can't just pass deps because object will change, need to make sure identity is constant

```ts
// Can't just use this any more
const retrieveDeps = asyncChoices.useRetrieveDeps({});
// We know we need pagiantor so pull that out. Making this entirely generic
// (eg. for other deps that might be in asyncChoices) is more work 
const { paginator } = retrieveDeps;
const resolve = useCallback(
    props => asyncChoices.retrieve(props, { paginator } )),
    [asyncChoices , paginator]
);
const selected = useAsyncValue({
     ids: value,
    resolve,
});
```

js-packages/@prestojs/rest/src/useAsyncLookup.ts
levic: This is an artifact of the doc generator isn't it? there's no way to nest the error/result section so that the other fields only display once?

davecoates: could update the doc renderer to handle it

js-packages/@prestojs/util/src/useAsync.ts
davecoates: added

js-packages/@prestojs/rest/src/CursorPaginator.ts
levic: Am not convinced that every cursor is going to be able to provide that, but if DRF does then I'm happy with it for now

js-packages/@prestojs/rest/src/useAsyncValue.ts
```diff
@@ -0,0 +1,249 @@
...
+    props: UseAsyncValuePropsMulti<T, U>
+): UseAsyncValueReturn<T[]>;
+export default function useAsyncValue<T, U extends Id>(
+    props: CommonProps<T, U> & {
+        id?: U | null;
+        ids?: U[] | null;
+        resolve: (idOrIds: U | U[]) => Promise<T[] | T>;
+    }
+): UseAsyncValueReturn<T | T[]> {
+    const { id, ids, getId, resolve, trigger } = props;
```
levic: Wasn't `trigger` going to get a default?

js-packages/@prestojs/rest/src/CursorPaginator.ts
davecoates: yeah it doesn't really matter - if they don't support it it can just return `true` always

it's just useful for UI to know when to show a 'more' button (so worst case is it just shows the button always)

levic (changes requested):

js-packages/@prestojs/rest/src/useAsyncLookup.ts
levic: Would prefer this if possible -- if we have tools (TS) that can catch errors then we should use them

(I presume another card for that?)

js-packages/@prestojs/util/src/useAsync.ts
levic: `comparisonFn` is a reference into a global table `comparisonByTrigger` -- it should be stable between renders

levic: edit: or do you mean that it's the existences of `comparisonFn` that is needed, and that the `One` part of `useMemoOne` is actually irrelevant?

davecoates: yes but `useMemo` doesn't support a custom comparison - we need `useMemoOne` (or something like it) for that

js-packages/@prestojs/rest/src/useAsyncValue.ts
```diff
@@ -0,0 +1,256 @@
...
```
levic: This trigger is missing a default

davecoates: fixed

js-packages/@prestojs/rest/src/useAsyncLookup.ts
davecoates: i've added the type changes, doc rendering changes later

js-packages/@prestojs/util/src/useAsync.ts
levic: Is this because the function changing doesn't actually trigger a rerender?

js-packages/@prestojs/rest/src/useAsyncValue.ts
```diff
@@ -0,0 +1,232 @@
...
+     * You can use this to avoid resolving data that already exists. If dealing with
+     * ViewModel instances you can use it with [useViewModelCache](doc:useViewModelCache),
+     * eg.
+     *
+     * ```js
+     * const existingValues = useViewModelCache(User, cache => cache.get(id, fieldNames));
+     * const { value } = useAsyncValue({ id, existingValues, resolve: fetchUser });
+     * ```
+     */
+    existingValues?: T[];
```
levic, line 32: I never asked this, but why is this an array rather than a mapping/dict?

If you had a large number of records this would become quite slow, or is the point that you shouldn't be passing in that many records anyway because it's unlikely you really have that many dependencies and it will cause unnecessary rerenders?

davecoates: Main reason is for ease of use with `useAsyncLookup` (you can pass the result of that straight through as `existingValues`). You'll also mostly have an array so passing a mapping/dict requires you to transform it to an object anyway (ie. you'd have to iterate over it all once anyway).

The other reason is compound keys wouldn't work as keys in either Map or object (would have to be stringified first)

PR #67, opened 2020-06-24
Feature/async choices 2

davecoates:


.isort.cfg
```diff
@@ -17,3 +17,4 @@ known_standard_library=
 known_third_party=
 	djrad
 	djrad_rest
+	allianceutils
```
levic: This doesn't seem to be referenced anywhere in the diff -- was it a bug fix or an accidental commit?

django-testbed/django-root/xenopus_frog/views.py
```diff
@@ -30,18 +32,38 @@ class Meta:
...
+class ModelIdFilter(ModelMultipleChoiceFilter):
+    def filter(self, qs, value):
+        return qs.filter(id__in=[record.id for record in value]) if value else qs
+
+
 class UserFilterSet(filters.FilterSet):
     first_name = filters.CharFilter(lookup_expr="icontains")
     last_name = filters.CharFilter(lookup_expr="icontains")
+    keywords = filters.CharFilter(method="filter_keywords")
+    ids = ModelIdFilter(queryset=User.objects.all())
```
levic: For consistency that should be the first field listed

.isort.cfg
davecoates: was causing CI to fail - has since been fixed in master

js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
```diff
@@ -46,19 +46,20 @@ export default function getWidgetForField<FieldValue, T extends HTMLElement>(
     field: Field<FieldValue>
 ): FieldWidgetType<FieldValue, T> | [FieldWidgetType<FieldValue, T>, object] | null {
     // Couldn't work out what to type this as so field.constructor was accepted
-    const widget: FieldWidgetType<any, any> | null | undefined = field.choices
-        ? choicesMapping.get(field.constructor.name) || mapping.get(field.constructor.name)
-        : mapping.get(field.constructor.name);
+    const widget: FieldWidgetType<any, any> | null | undefined =
+        field.choices || field.asyncChoices
+            ? choicesMapping.get(field.constructor.name) || mapping.get(field.constructor.name)
+            : mapping.get(field.constructor.name);
```
levic, line 52: This is `(field.choices || field.asyncChoices) ? ... : `? -- it might be worth putting brackets around that to make it a bit clearer what the order of precedence is (or maybe it's just me not seeing enough of these JS patterns)

js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
```diff
@@ -46,19 +46,20 @@ export default function getWidgetForField<FieldValue, T extends HTMLElement>(
     field: Field<FieldValue>
 ): FieldWidgetType<FieldValue, T> | [FieldWidgetType<FieldValue, T>, object] | null {
     // Couldn't work out what to type this as so field.constructor was accepted
-    const widget: FieldWidgetType<any, any> | null | undefined = field.choices
-        ? choicesMapping.get(field.constructor.name) || mapping.get(field.constructor.name)
-        : mapping.get(field.constructor.name);
+    const widget: FieldWidgetType<any, any> | null | undefined =
+        field.choices || field.asyncChoices
```
levic, line 53: This is `(field.choices || field.asyncChoices) ? ... : ... `? -- it might be worth putting brackets on line 50 to make it a bit clearer what the order of precedence is (or maybe it's just me not seeing enough of these JS patterns). Assuming prettification doesn't remove them.

js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
```diff
@@ -72,9 +73,10 @@ export default function getWidgetForField<FieldValue, T extends HTMLElement>(
     // if no match can be found check prototypes
     let f = Object.getPrototypeOf(field.constructor);
     do {
-        const widgetF: FieldWidgetType<any, any> | null | undefined = field.choices
-            ? choicesMapping.get(f.name) || mapping.get(f.name)
-            : mapping.get(f.name);
+        const widgetF: FieldWidgetType<any, any> | null | undefined =
+            field.choices || field.asyncChoices
```
levic, line 80: As above re parentheses

js-packages/@prestojs/util/src/comparison.ts
```diff
@@ -17,7 +18,7 @@ function equalityCustomizer(a: any, b: any): boolean | undefined {
     if (supportsEqualityInterface(b)) {
         return b.isEqual(a);
     }
-    // undefined return means shallowequal does normal shallow comparison
+    // undefined return means shallowequal or isEqualWith does normal shallow/deep comparison
```
levic: Grammar there is off

django-testbed/frontend/src/models/User.js
```diff
@@ -1,19 +1,27 @@
...
 import namedUrls from '../namedUrls';
 
 import BaseUser from './generated/BaseUser';
 
 function transformAndCacheUser(data) {
+    if (data.results && data.count != null) {
+        throw new Error(
+            'Response is paginated but paginator was not defined. Do you need to use PaginatedEndpoint instead?'
+        );
+    }
```
levic: For a normal app you'd presumably either know whether or not pagination was going to be present on the backend.

Is this something that should perhaps be rolled into presto core and printed as a console warning in dev?

django-testbed/frontend/src/models/User.js
```diff
@@ -24,7 +32,102 @@ function transformAndCacheUser(data) {
...
+                // default to pageNumber pagination if not specified
+                query: { ...query, paginationType: query.paginationType || 'pageNumber' },
+            });
+        },
+    }),
+    create: new Endpoint(namedUrls.get('users-list'), {
+        transformResponseBody: transformAndCacheUser,
+        method: 'post',
+    }),
+};
```
levic, line 35: The `endpoints` definition above we'd put in the code generator?

js-packages/@prestojs/ui-antd/src/widgets/SelectAsyncChoiceWidget.tsx
```diff
@@ -0,0 +1,468 @@
...
+type ValueType<T> = T[] | T | null;
+
+type SelectInputProps<T> = InputProps<ValueType<T>, HTMLSelectElement> & {
+    // Types in antd require event. Our types don't because final-form doesn't.
+    onBlur: (event: React.FocusEvent<HTMLSelectElement>) => void;
+    onFocus: (event: React.FocusEvent<HTMLSelectElement>) => void;
+};
+
+/**
+ * @expand-properties Most [Select](https://next.ant.design/components/select/) props in addition to the below. Note the
```
levic, line 18: That link is broken.

js-packages/@prestojs/ui-antd/src/widgets/SelectAsyncChoiceWidget.tsx
```diff
@@ -0,0 +1,468 @@
...
+
+type SelectAsyncChoiceWidgetValue = number | string;
+
+// Pagination is handled by rendering a button within a Select.Option that triggers a fetch. This
+// is the value used for that option. We can can prevent clicks on the option in case the button
+// fails to stop propagation and prevent an invalid change from occurring
+const NEXT_PAGE_VALUE = '__nextpage';
+
+/**
+ * A [Select](https://next.ant.design/components/select/) widget that handles [async choices](doc:AsyncChoicesInterface)
```
levic, line 256: Comments re link & ant coupling as above

js-packages/@prestojs/ui-antd/src/widgets/SelectAsyncChoiceWidget.tsx
levic: How tightly coupled is this to ant?

More specifically, this is the antd widget implementation so I would expect it to be highly dependent on ant, but if I wanted to use another widget lib then could I potentially to see a quite different async choices interface?

django-testbed/frontend/src/models/User.js
```diff
@@ -24,7 +32,102 @@ function transformAndCacheUser(data) {
...
+                        'email',
+                        'region',
+                    ]),
+                };
+            },
+            async retrieve(value, deps) {
+                return (await endpoints.list.execute({ query: { ids: value }, ...deps })).result;
+            },
+        }),
+    }),
```
levic, line 106: There seems to be a lot more boilerplate than I would have expected for this -- is this all required or is it testing every option?

Do you have a minimal example for an async choices that would be a lookup into another table? eg a list of Organisations, perhaps filtered by what the user is allowed to see.

js-packages/@prestojs/util/src/Labeled.ts
```diff
@@ -0,0 +1,110 @@
+/**
+ * Interface for items that provide a plain text label. Implementing this can save you having to pass
+ * explicit functions to label an item in other parts of the system (eg. for [AsyncChoices](doc:AsyncChoices))
```
levic, line 3: Am I able to read AsyncChoices documentation or only after it's merged?

js-packages/@prestojs/util/src/Labeled.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * See [isPlainLabeled](doc:isPlainLabeled) and [getPlainLabel](doc:getPlainLabel).
+ *
+ * @extract-docs
+ * @menu-group Labeled
+ */
+export interface PlainLabeled {
```
levic: Is Plain vs Rich used elsewhere in react?

It makes sense but it's just terminology I haven't seen before. (TextLabelled vs NodeLabelled maybe)?

Beyond that, if a string and a Node are both renderable by React why do we need the distinction?

levic: Edit: is this because some ant components insist on strings instead of renderable nodes?

js-packages/@prestojs/util/src/identifiable.ts
```diff
@@ -2,50 +2,82 @@ type SingleId = string | number;
...
+    /**
+     * The id for this item
+     */
+    _pk: Id;
+}
+
+/**
+ * Defines a property `id` to uniquely identify the item
+ */
+interface IdentifiableId {
```
levic: Why does this exist? If we already have `IdentifiablePk` then shouldn't that be the thing used to access the PK, and `id` is just the site-specific implementation of using `id` as the concrete PK?

If another backend was to use `Id` or `MyModel_Id` then wouldn't that require changes to presto?

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
```diff
@@ -0,0 +1,242 @@
...
+     * Hook that returns dependencies for listing choices. This is useful to store state
+     * for things like pagination. The value returned here is passed to `list`.
+     *
+     * What this function is passed depends on the implementation but when used with
+     * [useAsyncChoices](doc:useAsyncChoices) it will be passed `value` (the currently
+     * selected value), `query` (the query object eg. to filter results with) and `listOptions`
+     * (any additional options passed on the `listOptions` prop to `useAsyncChoices`).
+     * Official presto widgets all use `useAsyncChoices` and so will use these parameters.
+     */
+    useListProps(args: any): any;
```
levic, line 61: Can you give an example here of what a "dependency" is? How does it differ from the actual data itself?

eg if I have a list of organisations & their countries then is this supposed to query the country list or is that supposed to be a join that is going as part of the data returned by `list()` ?

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * Function to resolve a list of choices based on the provided params.
```
levic, line 63: `params` is where an autocomplete keyword search string would fit in? How would that get injected given that `useListProps()`is passed straight in?

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
```diff
@@ -0,0 +1,242 @@
...
+     * passed through to `useAsyncValue` as the list of items it can resolve existing values from.
+     *
+     * What this function is passed depends on the implementation but when used with
+     * [useAsyncChoices](doc:useAsyncChoices) it will be passed `id` if there's a current value
+     * and it's not an array, `ids` if there's a current value and it is an array, `existingValues` which is
+     * the values returned by `list` (may be null if `list` not yet called) and `retrieveOptions`
+     * (any additional options passed on the `retrieveOptions` prop to `useAsyncChoices`).
+     * Official presto widgets all use `useAsyncChoices` and so will use these parameters.
+     */
+    useRetrieveProps(args: any): any;
```
levic, line 86: Same comment as `useListProps` above

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
```diff
@@ -0,0 +1,242 @@
...
+     * Function to resolve specific values. This is used to know how to render the label for a value(s).
+     *
+     * The first parameter is the value to retrieve (will be an array when `multiple` is true).
+     *
+     * `deps` is the value returned by `useRetrieveProps`.
+     */
+    retrieve(
+        value: ValueType[] | ValueType,
+        deps?: any
+    ): Promise<typeof value extends ValueType[] ? ItemType[] : ItemType>;
```
levic, line 98: Same comment as `list` above

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
```diff
@@ -0,0 +1,242 @@
...
+     * Get a label for an item
+     */
+    getLabel(item: ItemType): React.ReactNode;
+
+    /**
+     * Return label to use when an item can't be found. This can be used by widgets to control
+     * what is rendered when an item for a value cannot be found (eg. when it's deleted or
+     * when it's loading). The exact details of how this is used depend on the widget.
+     */
+    getMissingLabel(value: ValueType): React.ReactNode;
```
levic, line 119: Is this intended to handle `null` values too or would the Widget handle that?

levic (changes requested):

django-testbed/frontend/src/models/User.js
davecoates: This is obsoleted by #68 which removes this check (because it's dependant on how pagination is implemented)

I couldn't think how to do the check generically so left it out but may be worth revisiting. Possible a catch all on errors that says something like 'There was a problem transforming response data. Is the response paginated?' if it fails on an unpaginated Endpoint

django-testbed/frontend/src/models/User.js
davecoates: If not initially then long term I think yes

django-testbed/frontend/src/models/User.js
davecoates: There's 3 fields defined here to cover different use cases. `referredBy` is the simple use case:

```
referredBy: new IntegerField({
   asyncChoices: new AsyncChoices(asyncChoicesOptions),
 }),
```

in terms of reducing `asyncChoicesOptions` it's fairly generic so I imagine we'll have `ViewModelAsyncChoices` or something like that that does that (may be default for foreign key fields?)

js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
davecoates: no can do, prettier removes them

davecoates: https://github.com/prettier/prettier/issues/3805

js-packages/@prestojs/ui-antd/src/widgets/SelectAsyncChoiceWidget.tsx
davecoates: Reasonably specific to how antd select works. The usage of async choices is specific to ant but the AsyncChoices interface is generic (or at least that's the goal)

js-packages/@prestojs/util/src/Labeled.ts
davecoates: I haven't seen `Rich` or `Plain`

The reason for the distinction is really for when you want to use string interpolation or other text transformation vs just render something

js-packages/@prestojs/util/src/identifiable.ts
davecoates: I think it was really because `id` is the intuitive thing if implementing `Identifiable` it's just that it's not suitable for our use case (although would work most of the time). That's where I started from anyway but didn't want to force `ViewModel` to conform to this so added `_pk` as the support. 

I guess we could just change it to `_pk` it just sounds a bit more specific to a database implementation

(this isn't new behaviour in this PR BTW - I just refactored how the type was defined to make it easier to work with internally)

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
davecoates: oops I changed the name from `useListDeps` to `useListProps` but still use dependencies in the comment.

The props returned here is anything required by the eg. widget or hooks that use the async choices. So pagination is the common thing:

```ts
useListProps() {
  const paginator = usePaginator(endpoints.list);
  return { paginator };
}
```

It also can be used to transform the props passed in - so if you wanted to change how the keywords from a generic widget are transformed before they are passed through to the async function (eg. the endpoint) you can do this here without having to have a custom widget.

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
davecoates: See https://github.com/prestojs/prestojs/pull/67/files#diff-a1b1883790df4d0587b34a90efa4a851R120

You can return it from `listProps` otherwise it's passed straight through in `useAsyncChoices`

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
davecoates: Same kind of deal - most obvious usage I had for it was reading from cache always for existing values to avoid unnecessary fetches, eg https://github.com/prestojs/prestojs/pull/67/files#diff-013e735fdb71bd7fe985d673e4eb416cR62

js-packages/@prestojs/viewmodel/src/fields/AsyncChoices.ts
davecoates: No this is just for if a value is present but no label can be found

davecoates: (widget handles it - there's a default on the widget or can be specified as a prop)

js-packages/@prestojs/util/src/Labeled.ts
davecoates: Normally yes but currently it's failing to build

js-packages/@prestojs/util/src/Labeled.ts
levic: Can we leave it up to the dev then whether it returns a string or a Node?

If they try to do string interpolation on a Node then they'll get an error, but they were the one who chose to return a Node so they can fix it.

davecoates: You wouldn't get an error would just end up with a string like "Welcome [object Object]"

The intention was to provide different guarantees

Rich - suitable to be rendered to screen 
Plain - suitable to be used in an alert box, fill a text box, send via API call etc

You don't think there's advantage in separating them?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -651,6 +657,11 @@ export default function viewModelFactory<T extends FieldsMapping>(
                 return this[pkFieldName];
             },
         },
+        _key: {
+            get(): PrimaryKey {
+                return this._pk;
```
levic, line 686: Is `_pk` still needed at all? Can we just replace it entirely with `_key`?

levic (commented):
Looks good other than the question about _key / _pk: is _pk there just because of history or does it have a distinct meaning from _key? (esp does ViewModel really need both?)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: Mainly because it was already there and used throughout + parallel to django's `record.pk`

i guess it would make sense to remove it and just use _key

js-packages/@prestojs/util/src/Labeled.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Get a plain text label from an item
+ *
+ * If does not implement TextLabeled an error is thrown. To conditional call this check if labeled first
+ * with `isTextLabeled`.
+ *
+ * @extract-docs
+ * @menu-group Labeled
+ */
+export function getPlainLabel(item: any): string {
```
levic: That should be `getTextLabel` for consistency with naming above

PR #68, opened 2020-06-29
feature: ViewModelEndpoint

### LINKED ISSUE #32: ViewModelEndpoint
Handle caching of data transparently 

eg. currently can be achieved like:

```js
function transformAndCacheTimecard(data) {
    if (Array.isArray(data)) {
        const tasks = data.map(datum => new Timecard(datum));
        Timecard.cache.addList(tasks);
        return tasks;
    }
    const user = new Timecard(data);
    Timecard.cache.add(user);
    return user;
}

export default class Timecard extends BaseTimecard {
   
    static endpoints = {
        create: new Endpoint(namedUrls.get('timecard-list'), {
            transformResponseBody: transformAndCacheTimecard,
            method: 'post',
        }),
    };

 }
```

davecoates:


js-packages/@prestojs/rest/src/__tests__/ViewModelEndpoint.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        },
+        {
+            id: 2,
+            firstName: 'Frodo',
+            lastName: 'Baggins',
+        },
+        {
+            id: 3,
+            firstName: 'Gandalf',
+            lastName: '',
```
levic, line 40: The ~Grey~ White!

js-packages/@prestojs/rest/src/ViewModelEndpoint.ts
```diff
@@ -0,0 +1,111 @@
...
+    return record;
+}
+
+type ViewModelMapping = ViewModelConstructor<any> | Record<string, ViewModelConstructor<any>>;
+type ViewModelMappingDef = ViewModelMapping | (() => ViewModelMapping | Promise<ViewModelMapping>);
+
+/**
+ * Define an endpoint for a ViewModel. The response is transformed according
+ * to `viewModelMapping`. If the endpoint returns a single record or a list
+ * of the same record type you can pass a ViewModel class as the parameter:
```
levic: This was a bit confusing, I've tried to rewrite it to make it clearer. Feel free to fix/alter anything that I've got wrong/made less clear (in particular, the fact that arrays vs individual objects is handled automatically)

```
/**
 * Define an endpoint for a ViewModel. The response is transformed according
 * to `viewModelMapping`.
 *
 * The simplest form of a mapping is to a Model:
 *
 * ```js
 * getUser = new ViewModelEndpoint(new UrlPattern('/users/:id/'), User);
 * ```
 *
 * If an element is a response is an array then it will transparently be treated as a list
 * of objects of the mapping type and the transformation function will be invoked for each
 * element one by one:
 * 
 * ```js
 * // Response is a single User instance:
 * const userRetrieve = new ViewModelEndpoint(new UrlPattern('/users/:id/'), User);
 * // Response is an array of User instances
 * // (declaration is the same but the response handler will treat it differently)
 * const userList = new ViewModelEndpoint(new UrlPattern('/users/'), User);
 * ```
 *
 * If the response is an object mapping different models you can specify how each
 * key is transformed. Each response value will be treated as an array/individual object
 * automatically:
 *
 * ```js
 * new ViewModelEndpoint(new UrlPattern('/users/'), {
 *     users: User,
 *     bookings: Booking,
 * });
 * ```
 *
 * Dot notation is also supported for nested objects (again each element may be a
 * single object or an array of objects)
 *
 * ```js
 * new ViewModelEndpoint(new UrlPattern('/users/'), {
 *     "records.users": User,
 *     "records.bookings": Booking,
 * });
 * ```
 *
 * The mapping can optionally be a function that returns a mapping or returns a promise that
 * resolves to a mapping. This is useful for dealing with dynamic imports or returning a class
 * that isn't yet defined.
 *
 * ```js
 * new ViewModelEndpoint(new UrlPattern('/users/'), () => {
 *   const Booking = (await import('./Booking')).default;
 *   return {
 *     "records.users": User,
 *     "records.bookings": Booking,
 *   }
 * });
 * ```
 *
 * If the endpoint is paginated use [PaginatedViewModelEndpoint](doc:PaginatedViewModelEndpoint).
 *
 * TODO: Add note about how foreign key data is cached once we have worked that out
 * TODO: Convention for deleting an item?
 *
 * @extract-docs
 */
```

levic: To clarify: if I have an object that contains a nested child object, the parent object transformation function still gets passed the full unabridged data for that object (including the raw data for the child object)?

levic: To clarify: the Promises form is only for defining the mapping? If I had a mapping function that returned a Promise that resolves to the mapped data then that wouldn't work? (I don't think it should, but when I initially saw that you could use Promises that's what I assumed it would be for)

davecoates: > To clarify: if I have an object that contains a nested child object, the parent object transformation function still gets passed the full unabridged data for that object (including the raw data for the child object)?

Correct. Typically though I don't think this would be useful - you'd use foreign key fields to handle those automatically for you

> To clarify: the Promises form is only for defining the mapping? If I had a mapping function that returned a Promise that resolves to the mapped data then that wouldn't work? (I don't think it should, but when I initially saw that you could use Promises that's what I assumed it would be for)

Yes that's correct in terms of the intention here but technically both will work as this transformation functions runs as part of a promise chain.. so returning a promise from the function itself will still work (can't think of a case when you'd actually need that though)

davecoates: Thanks - I've updated this now

davecoates:
Merged into #67

PR #74, opened 2020-07-16
Feature/related viewmodel field

### LINKED ISSUE #49: Foreign key support (related view models)
For foreign keys they are an extension of #48  - eg. `asyncChoices` is something on the base field that can be provided for Foreign Key fields. The main usage in presto for foreign keys would be for selecting data, eg. cache usage might be

```js
// Get user from cache along with all subscriptions and the product name
cache.get(User, ['firstName', 'lastName', 'email', 'subscriptions.product.name'])
```

But also need it in data fetching. Simplest implementation without trying to replicate graphql is just to bring back the batch API from djrad1 and have the traversal just batch up all the necessary API calls to the backend. I think this would become part of the Endpoint interface via specifying fields to fetch (currently this isn't part of the interface at all).

So maybe the interface for a FK field would be

```
new RelatedManyViewModel({
   to: () => import('./UserSubscription'),
   resolve: (viewModel, ids, fieldNames) => viewModel.endpoints.list(...),
   remoteChoices: {....},
})
```

`asynChoices` handles how values are filtered and shown in a widget so all the `RelatedViewModel` or `RelatedManyViewModel` needs to do is resolve 1 or many objects based on id. Caching implementation can then handle resolving cached objects across multiple caches.

davecoates:
Closes #49

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
+++ b/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
@@ -365,20 +365,33 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
 
     /**
      * Adds a placeholder entry for `fieldNames` so that changes to that cache key are detected
-     * (eg. when a superset of fields are changed). Changes to a subset of fields are detected
-     * by their existence in either listener cache or record cache. This handles almost all cases
-     * but as we support traversing related caches there is the case where some part of the chain
-     * legitimately has a null value on the sourceFieldName and so the related record can't be
-     * returned but it's still a valid record.
+     * (eg. when a superset of fields are changed).
      *
-     * For example if you try to retrieve a record with `get(1, ['user', 'group', 'owner'])`
-     * but the `groupId` is null (ie. group is actually optional) then this should still return
-     * a record as it exists it's just that the `RelatedViewModelField` field isn't set. This won't
-     * be detected based on entries in cache keys or listeners keys as we can't extract the nested
-     * field paths as they aren't set. In this case we know what fields were requested (they were
-     * specified in the `get` call) and so we use this function to add a null entry in the cache
-     * purely so the key exists and can be returned.
-     * @param fieldNames
+     * Changes to a subset of fields are detected by their existence in either the listener cache
+     * or record cache. If a record is retrieved from the cache on field names that has no current
+     * entry but there exists a record with the superset of fields then a new cache Record entry
+     * is created at that point in time and returned immediately.
+     *
+     * This works fine in most cases however if there's a null RelatedField then keys corresponding
+     * to fields nested via the FK won't yet exist in the cache and will not be updated. This is because
+     * when the record is cached the keys are extracted from the assigned fields - but if the RelatedField
+     * is null then the nested fields cannot be inferred and so the cache key will differ from the
+     * requested key.
+     *
+     * As an example consider `cache.get(1, ['username', ['group', 'name']])`
+     *  - Here there's a related field `group` which in turn has a related field `owner`.
+     *  - The source field for `group` is `groupId` and is null.
+     *  - The requested key does not exists we do have a superset of
+     *    `['email', 'username', ['group', 'name']]`
+     *  - `addKeyPlaceholder(['username', ['group', 'name']])` is called to ensure that there is
+     *    guaranteed to be a cache entry for the requested key.
+     *    - If there is no cache entry present it will set it to null.
+     *  - We set the cache entry for `['username', ['group', 'name']])` to
+     *    `{username: 'dev', groupId: null}`
+     *  - The act of setting this cache entry also triggers an update on the `['username', 'groupId']`
+     *    key which is identified as having a subset of fields
+     *  - This ensure that updates to partial records involving both `group` and `groupId` are kept
+     *    in sync
      */
     addKeyPlaceholder(fieldNames: FieldPath[]): void {
         fieldNames = expandRelationFieldPaths(this.viewModel, fieldNames as FieldPath[]);
--- a/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
+++ b/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
@@ -15,6 +15,29 @@ import {
     ViewModelInterface,
 } from './ViewModelFactory';
 
+// Controller for conditionally enabling listeners. We don't want to fire listeners anytime
+// something changes in the cache as there's lots of changes that happen internally for keeping
+// the cache in sync that aren't specifically a change consumers care about
+let listenersEnabled = false;
+
+/**
+ * Any changes that occur within the passed function will trigger listeners
+ */
+function withEnableListeners<T>(run: () => T): T {
+    if (listenersEnabled) {
+        return run();
+    }
+    listenersEnabled = true;
+    try {
+        const r = run();
+        listenersEnabled = false;
+        return r;
+    } catch (err) {
+        listenersEnabled = false;
+        throw err;
+    }
+}
+
 /**
  * Points to a record that is cached already. The purpose of this is to have a single object
  * in memory and only clone it as needed (eg. a single instance of the RecordPointer is stored
@@ -128,6 +151,9 @@ type AllChangesListener = () => void;
 // ['a', 'b', 'c'] becomes 'a⁞b⁞c'
 const CACHE_KEY_FIELD_SEPARATOR = '⁞';
 
+/**
+ * Must be kept in sync with `reverseCacheKey`
+ */
 function getFieldNameCacheKey(
     fieldNames: readonly FieldPath[],
     viewModel: ViewModelConstructor<any>
@@ -204,6 +230,8 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
 
     /**
      * Return the key to use into `cache` for the specified field names
+     *
+     * Must be kept in sync with `reverseCacheKey`
      */
     private getCacheKey(
         fieldNames: readonly (keyof ViewModelClassType['fields'])[] | FieldPath[]
@@ -213,6 +241,8 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
 
     /**
      * Take a cache key generated with `getCacheKey` and return the list of fields
+     *
+     * Must be kept in sync with `getCacheKey`/`getFieldNameCacheKey`
      */
     private reverseCacheKey(fieldsKey: string): FieldPath[] {
         return fieldsKey.split(CACHE_KEY_FIELD_SEPARATOR).map(fieldName => {
@@ -236,7 +266,8 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
      * @private
      */
     private getRelationFields(
-        fieldNames: FieldPath[]
+        fieldNames: FieldPath[],
+        viewModel = this.viewModel
     ): { relations: Record<string, FieldPath[]>; nonRelationFieldNames: string[] } {
         const relations: Record<string, FieldPath[]> = {};
         const nonRelationFieldNames: string[] = [];
@@ -248,16 +279,18 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
                 } else {
                     relations[fieldName[0]].push(fieldName.slice(1));
                 }
-            } else if (this.viewModel.fields[fieldName] instanceof RelatedViewModelField) {
+            } else if (viewModel.fields[fieldName] instanceof RelatedViewModelField) {
                 relations[fieldName] = relations[fieldName] || [];
-                nonRelationFieldNames.push(this.viewModel.fields[fieldName].sourceFieldName);
+                nonRelationFieldNames.push(viewModel.fields[fieldName].sourceFieldName);
             } else {
-                nonRelationFieldNames.push(fieldName);
+                if (!viewModel.pkFieldNames.includes(fieldName)) {
+                    nonRelationFieldNames.push(fieldName);
+                }
             }
         }
         for (const [relationName, relationFieldNames] of Object.entries(relations)) {
             if (relationFieldNames.length === 0) {
-                const relation = this.viewModel.fields[relationName];
+                const relation = viewModel.fields[relationName];
                 // If field names for relation haven't been specified default to all non-relation fields
                 relationFieldNames.push(
                     ...relation.to.fieldNames.filter(fieldName => {
@@ -307,7 +340,9 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
                 } else {
                     ret = this.cache.set(key, value);
                 }
-                listeners.forEach(cb => cb(before as InstanceType<ViewModelClassType>, value));
+                if (listenersEnabled) {
+                    listeners.forEach(cb => cb(before as InstanceType<ViewModelClassType>, value));
+                }
             }
         } else {
             if (value === null) {
@@ -333,14 +368,16 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
      * (eg. when a superset of fields are changed). Changes to a subset of fields are detected
      * by their existence in either listener cache or record cache. This handles almost all cases
      * but as we support traversing related caches there is the case where some part of the chain
-     * legitimately has a null value and so the related record can't be returned but it's still
-     * a valid record. For example if you try to retrieve a record with `get(1, ['user', 'group', 'owner'])`
+     * legitimately has a null value on the sourceFieldName and so the related record can't be
+     * returned but it's still a valid record.
+     *
+     * For example if you try to retrieve a record with `get(1, ['user', 'group', 'owner'])`
      * but the `groupId` is null (ie. group is actually optional) then this should still return
-     * a record as it exists it's just that the relations aren't set. This won't be detected based
-     * on entries in cache keys or listeners keys as we can't extract the nested field paths as
-     * they aren't set. In this case we know what fields were requested (they were specified in the
-     * `get` call) and so we use this function to add a null entry in the cache purely so the key
-     * exists and can be returned.
+     * a record as it exists it's just that the `RelatedViewModelField` field isn't set. This won't
+     * be detected based on entries in cache keys or listeners keys as we can't extract the nested
+     * field paths as they aren't set. In this case we know what fields were requested (they were
+     * specified in the `get` call) and so we use this function to add a null entry in the cache
+     * purely so the key exists and can be returned.
      * @param fieldNames
      */
     addKeyPlaceholder(fieldNames: FieldPath[]): void {
@@ -385,7 +422,9 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
                                 record = record.clone(assignedFieldNames);
                             }
                             if (record) {
-                                if (!after.isEqual(record[fieldName])) {
+                                if (!after) {
+                                    this.cleanupKey(fieldsKey);
+                                } else if (!after.isEqual(record[fieldName])) {
                                     const newRecord = new this.viewModel({
                                         ...record._data,
                                         [fieldName]: after,
@@ -403,6 +442,71 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
     }
 
     /**
+     * Given an existing cached record `record`, a cache key `key` and the
+     * fields `assignedFieldNames` on `record` create a new record for `key` if it is
+     * a subset of `record`.
+     * @private
+     */
+    private createRecordForSubKey(
+        record: InstanceType<ViewModelClassType> | RecordPointer<ViewModelClassType>,
+        key: FieldNameCacheKey,
+        assignedFieldNames: FieldPath[]
+    ): InstanceType<ViewModelClassType> | RecordPointer<ViewModelClassType> | null {
+        const fieldNames = flattenFieldPath(assignedFieldNames);
+        const cacheFieldNames = this.reverseCacheKey(key);
+        const { nonRelationFieldNames: otherNonRelationFieldKeys } = this.getRelationFields(
+            fieldNames
+        );
+        const { relations, nonRelationFieldNames } = this.getRelationFields(cacheFieldNames);
+        if (isSubset(flattenFieldPath(cacheFieldNames), fieldNames)) {
+            if (record instanceof RecordPointer) {
+                record = record.record;
+            }
+            return new RecordPointer(this.cache.get(key), record);
+        }
+        if (
+            Object.keys(relations).length > 0 &&
+            isSubset(nonRelationFieldNames, otherNonRelationFieldKeys)
+        ) {
+            if (record instanceof RecordPointer) {
+                record = record.record;
+            }
+            const data = {};
+            for (const [fieldName, relationFields] of Object.entries(relations)) {
+                const relation = this.viewModel.fields[fieldName];
+                if (record._assignedFields.includes(relation.sourceFieldName)) {
+                    const id = record[relation.sourceFieldName];
+                    if (id == null) {
+                        data[fieldName] = null;
+                        data[relation.sourceFieldName] = null;
+                    } else {
+                        let relationRecord = relation.to.cache.get(id, relationFields);
+                        if (
+                            !relationRecord &&
+                            record._assignedFields.includes(fieldName) &&
+                            record[fieldName]
+                        ) {
+                            relation.to.cache.add(record[fieldName]);
+                            relationRecord = relation.to.cache.get(id, relationFields);
+                        }
+                        if (!relationRecord) {
+                            return null;
+                        }
+                        data[fieldName] = relationRecord;
+                    }
+                } else {
+                    return null;
+                }
+            }
+            return new this.viewModel({
+                ...pick(record._data, [...nonRelationFieldNames, ...record._model.pkFieldNames]),
+                ...data,
+            }) as InstanceType<ViewModelClassType>;
+        }
+        return null;
+    }
+
+    /**
      * Add a record to the cache based on the fields that are set on it.
      *
      * This will also update any cached entries for records that contain only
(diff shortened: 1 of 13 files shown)


js-packages/@prestojs/viewmodel/src/fields/Field.ts
```diff
@@ -264,6 +265,17 @@ export default class Field<T, ParsableType extends any = T> {
...
+    /**
+     * Called after fields are attached to a ViewModel.
+     *
+     * By default this does nothing but can be used by fields to attach extra properties or validate
+     * against the final view model (for example checking that another field does / does not exist).
+     */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    public contributeToClass(viewModel: ViewModelConstructor<any>): void {
+        // Do nothing by default
+    }
```
levic, line 321: This only gets called by the factory? Should mention that this will only be called by the factory and that the immediate child will simply inherit the contributed field.

Does anything special happen for deeper model inheritance?

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
```diff
@@ -0,0 +1,267 @@
...
+ * `to` can also be a a function that returns a Promise. This is useful to
+ * lazy load modules:
+ *
+ * ```js
+ * class Subscription extends viewModelFactory({
+ *   userId: new IntegerField(),
+ *   user: new RelatedViewModelField({
+ *       sourceFieldName: 'userId',
+ *       to: async () => {
+ *         const User = await import('./User').default;
```
levic, line 124: Do you have an example of how this would be useful?

I would have thought that if contributeToClass() is called when the factory is called then this is still going to trigger an import on page load, so you don't really get that much do you?

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
```diff
@@ -0,0 +1,267 @@
...
+ *   })
+ * }) {}
+ * ```
+ *
+ * **NOTE:** When you return a promise you have to call `resolveViewModel` on
+ * that field before it's usable:
+ *
+ * ```js
+ * await Subscription.fields.user.resolveViewModel()
+ * ```
```
levic, line 136: When would you expect this would be called? By a widget?

Also, I assume you can call it multiple times with no ill effects? Can you state this explicitly

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -493,6 +509,217 @@ function checkReservedFieldNames(fields): void {
...
+        if (typeof path === 'string') {
+            flattenedPath.push(path);
+        } else {
+            flattenedPath.push(path.join(separator));
+        }
+    }
+    return flattenedPath;
+}
+
+export function expandField(model: ViewModelConstructor<any>, fieldName: string): FieldPath[] {
```
levic: Missing a docstring

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -493,6 +509,217 @@ function checkReservedFieldNames(fields): void {
...
+    return [fieldName];
+}
+
+/**
+ * Thrown when attempting to access a field that does not exist on a ViewModel
+ */
+export class InvalidFieldError extends Error {}
+
+/**
+ * For the given field paths expand any relation fields to include the nested non-relation fields
```
levic, line 612: `the nested non-relation fields`

the meaning of that is unclear.

Do you mean fields on records accessible by traversing FK relationships?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -493,6 +509,217 @@ function checkReservedFieldNames(fields): void {
...
+ * ```
+ * [
+ *   ['user', 'group', 'name'],
+ *   ['user', 'group', 'ownerId'],
+ * ]
+ * ```
+ *
+ * Where 'user' is a foreign key to a model that has a foreign key on field 'group' with two fields.
+ *
+ * We exclude non-relation fields to avoid circular dependencies.
```
levic: How can a non-relation field create a circular dependency? Can't you only create circular references through FK relationships? (Although the code seems to match what you wrote)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -493,6 +509,217 @@ function checkReservedFieldNames(fields): void {
...
+                }
+                const field = currentModel.fields[fieldName];
+                if (!field) {
+                    throw new InvalidFieldError(
+                        `Invalid field ${fieldName} on model ${currentModel}`
+                    );
+                }
+                const isRelation = field instanceof RelatedViewModelField;
+                const isLast = i === path.length - 1;
+                if (i % 2 == 0 && !isLast && !isRelation) {
```
levic: What's the `i % 2` for? -- why do we only care about even numbered dotted paths?

Either a comment or an extra var, eg `isFoo = i % 2 == 0` (I don't understand what the condition actually being tested is)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -631,6 +861,29 @@ export default function viewModelFactory<T extends FieldsMapping>(
...
+                        data[field.sourceFieldName] !== assignedData[key]._pk
+                    ) {
+                        console.warn(
+                            `Related field ${
+                                field.name
+                            } was created from nested object that had a different id to the source field name ${
+                                field.sourceFieldName
+                            }: ${data[field.sourceFieldName]} !== ${assignedData[key]._pk}. ${
+                                assignedData[key]._pk
+                            } has been used for both.`
```
levic: The prettier fomatting on that is horrible -- maybe concat substrings together to make it easier to follow?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -927,6 +1237,56 @@ export default function viewModelFactory<T extends FieldsMapping>(
                 return Object.keys(this.fields).filter(name => !pkFieldNames.includes(name));
             },
         },
+        getField: {
```
levic, line 1272: How does `getField('foo')` differ from `fields['foo']` and when should one be used over the other?

I would have thought that they'd amount to the same thing but that doesn't appear to be the case?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -119,9 +128,27 @@ type AllChangesListener = () => void;
 // ['a', 'b', 'c'] becomes 'a⁞b⁞c'
 const CACHE_KEY_FIELD_SEPARATOR = '⁞';
 
-function getFieldNameCacheKey(fieldNames: readonly string[], excludeFields: string[]): string {
+function getFieldNameCacheKey(
```
levic, line 157: Without giving a detailed explanation (since this is basically opaque and self-contained to callers) can you give one or two examples of what this will generate?

levic: Should mention that this needs to be kept in sync with `reverseCacheKey()`

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -155,34 +182,91 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
+    ): string {
+        return getFieldNameCacheKey(fieldNames as FieldPath[], this.viewModel);
     }
 
     /**
      * Take a cache key generated with `getCacheKey` and return the list of fields
      */
-    private reverseCacheKey(fieldsKey: string): string[] {
-        return fieldsKey.split(CACHE_KEY_FIELD_SEPARATOR);
+    private reverseCacheKey(fieldsKey: string): FieldPath[] {
```
levic, line 247: Should mention that this needs to be kept in sync with `getCacheKey()`/`getFieldNameCacheKey()`

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -243,10 +324,84 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
 
+    /**
+     * Adds a placeholder entry for `fieldNames` so that changes to that cache key are detected
+     * (eg. when a superset of fields are changed). Changes to a subset of fields are detected
+     * by their existence in either listener cache or record cache. This handles almost all cases
+     * but as we support traversing related caches there is the case where some part of the chain
+     * legitimately has a null value and so the related record can't be returned but it's still
+     * a valid record. For example if you try to retrieve a record with `get(1, ['user', 'group', 'owner'])`
+     * but the `groupId` is null (ie. group is actually optional) then this should still return
+     * a record as it exists it's just that the relations aren't set. This won't be detected based
```
levic: I'm confused about this comment: "a record as it exists it's just that the relations aren't set"

If groupId is null then there are no Group or Owner records. What do you mean by "it exists"?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -299,84 +527,108 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
-            return recordOrPointer as PartialViewModel<
-                ViewModelClassType,
-                typeof fieldNames[number]
-            >;
+            return recordOrPointer;
         }
         return null;
     }
 
+    private cleanupKey(fieldsKey): void {
```
levic, line 645: What does this do? Purge a record from the cache? How does it differ from `delete()`?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -395,6 +647,101 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
+     */
+    batch<T>(run: () => T): T {
+        if (this.isActive) {
+            return run();
+        }
+        this.isActive = true;
+        try {
+            const r = run();
+            let pending = this.pendingNoBatch;
+            // TODO: :shrug:?
```
levic: ?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -395,6 +647,101 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
+                this.pendingNoBatch = new Map();
+                pending.forEach(([before, after], cb) => cb(before, after));
+                pending = this.pendingNoBatch;
+                // break;
+            }
+            this.pending.forEach(([before, after], cb) => cb(before, after));
+            this.pending = new Map();
+            this.pendingAll.forEach(cb => cb());
+            this.pendingAll = new Set();
+            // TODO:Should this move up above this.pending.forEach?
```
levic: Why would you want to do this? Until this point isn't it still active?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -471,6 +818,55 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
  * // [new User({ id: 3, name: 'Jayz' }), new User({ id: 4, name: 'Beeb' })] changed to [null, new User({ id: 4, name: 'Beeb' })]
  * ```
  *
+ * ## Field notation
+ *
+ * If a model has a [RelatedViewModelField](doc:RelatedViewModelField) the data for a related field
+ * can be retrieved using array notation:
+ *
+ * ```js
+ * // Fetch the 'name' field and the related 'group' record and it's 'label' field
```
levic: `it's` => `its`

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -471,6 +818,55 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
+ * ```
+ *
+ * You can combine the shorthand with array notation to get all fields and the specified deep relations:
+ * ```js
+ * ['name', 'group', ['group', 'owner', 'name']]
+ * // Equivalent to:
+ * ['name', ['group', 'label'], ['group', 'owner', 'name']]
+ * ```
+ *
+ * **NOTE:** When accessing a relation it's `sourceFieldName` is always included regardless
```
levic: `it's` => `its`

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -488,6 +885,33 @@ export default class ViewModelCache<
...
+     * ```js
+     * User.cache.addListener(listenerAll);
+     * User.cache.addListener(1, ['id', 'name'], listener);
+     * User.cache.addListenerList([1, 2], ['id', 'name'], listenerList);
+     * User.cache.batch(() => {
+     *   // This value won't appear in changes at all as it's replaced 2 lines down
+     *   User.cache.add({ id: 1, name: 'Bob', groupId: 1 });
+     *   User.cache.add({ id: 2, name: 'Sam', groupId: null });
+     *   User.cache.add({ id: 1, name: 'Bobby', groupId: 1 });
+     * });
```
levic, line 1000: What happens if you try to nest batches?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -395,6 +647,101 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
+            }
+        } else {
+            listener(before, after);
+        }
+    },
+    /**
+     * Start a batch. The passed function is called and any changes queued. Once function returns listeners
+     * will be dispatched. The value returned from `run` will be returned.
+     *
+     * If any error occurs in `run` then no listeners will be called.
```
levic, line 805: See comment later on: what is expected to happen if you try to nest batches?

(implementation doesn't seem to handle this, but even if it did then there's a question of when the notification should take place, esp if the same record is modified in both an inner and outer batch)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -540,39 +964,42 @@ export default class ViewModelCache<
...
+            if (!this.isInstanceOfModel(recordOrData)) {
+                if (isViewModelInstance(recordOrData)) {
+                    throw new Error(
+                        `Attempted to cache ViewModel of type ${recordOrData._model} in cache for ${
+                            this.viewModel
+                        }. ${
+                            isDev()
+                                ? ' If you are using hot loading this can cause instanceof checks to fail when the class is recreated - if this is the case a hard refresh should resolve the issue.'
+                                : ''
+                        }`
```
levic: Can you restructure that to work around prettier's indentation vomit

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -674,12 +1095,157 @@ export default class ViewModelCache<
...
+            }
+            if (!sourceFieldNames.includes(sourceFieldName)) {
+                sourceFieldNames.push(sourceFieldName);
+            }
+        }
+        // If record exists under fieldNames key already then return it
+        let record = recordCache.get(fieldNames as FieldPath[]);
+        if (record) {
+            return record;
+        }
```
levic: Can you chat to me about this? I would have thought that you'd never store it with the related record and always look the record based on source field(s) only, and then inject the record reference back in

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -674,12 +1095,157 @@ export default class ViewModelCache<
...
+                // We had an id but the underlying record isn't available
+                if (!relatedRecord) {
+                    return null;
+                }
+                relationData[relation.name] = relatedRecord;
+            }
+            if (Object.keys(relationData).length > 0) {
+                // This could exist as cachedFieldNames can be changed above - if so we can return the
+                // record now otherwise we must create a new cache entry for the record with resolved
+                // relations.
```
levic, line 1321: See comment above: do we need to cache this?

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
davecoates: contributeToClass is called on the field so it would only trigger that if you happened to access `to` in that function (we don't)

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
davecoates: Honestly don't know - was thinking just in the user app but I guess the widget could do this for us automatically and just return null (or use suspense?) until it's loaded 🤔

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: fields on a related model yes - but not recursive (hence the non-relation fields only)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: oops it's supposed to say 'exclude relation fields'

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: ummm yeah dunno what I was thinking there. it should just be everything except last element has to be a relation field

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: that example is the same

```
getField(['fk1', 'fk2', 'foo'])
```

is the difference (ie. you can traverse relations with getField)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: the relation could be null because either a) the related record isn't in the cache b) there's just no relation (ie. the id is set to null)

So I just mean it should be a cache hit - it exists in the cache despite not having those nested fields (ie. it can't have the nested fields because the relation is null)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: just a private internal method `delete` calls to remove cache entry and cleanup listeners (you'll see in `delete` it calls this function in 2 different code paths hence why it's a separate function).

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: This is the 'flush' phase... if any of this:

```
this.pending.forEach(([before, after], cb) => cb(before, after));
            this.pending = new Map();
            this.pendingAll.forEach(cb => cb());
            this.pendingAll = new Set();
```

triggered another callback... then they would be batched but lost as we set `this.pending = new Map()` afterwards.

I wasn't sure what I should do here but I think in practice it won't happen (it's all internal stuff - I added `this.pendingNoBatch ` purely for this reason)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: They all run in the same batch

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: the implementation handles it - batcher is just a singleton if you call it while it's already in a batch it just calls immediately:

```
if (this.isActive) {
    return run();
}
```

the notification happens at end of outer most batch

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: easier implementation this way & easier to guarantee the record is always the same memory location so `===` always passes

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: so that next time you retrieve it from the cache it's the same object + don't have to go through all the logic to work out if it should exist

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
levic: Please add a comment in one of those functions highlighting the difference

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
levic: Ah so when you say "the relation could be null", the FK is not null but the RelatedField field is?

Can you clarify that a bit more in the comments -- specifically when you say "the relation" it's not clear that you mean "the RelatedViewModelField field" and not necessarily the FK source field

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
levic: If not already there, can you double check that test cases confirm this works as expected

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
levic: "the notification happens at end of outer most batch" - that should be documented

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
levic: If someone updates the cache then doesn't it replace the record? (Or does it just update the fields on an existing record?)

If there's no test case for that already, can you make sure that replacing a record in the cache doesn't cause records that point to it to use the older (stale) record. ie pull back a User record with a related Group, then update the Group cache with a new Group record, then check that the User record correctly points to the new Group

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: It's already documented on the interface, see https://deploy-preview-74--prestojs.netlify.app/docs/viewmodel/viewmodelfactory/#method-getField

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: it replaces the record in the cache

yeah there's lots of tests around that

js-packages/@prestojs/viewmodel/src/fields/Field.ts
davecoates: It gets called for every class anywhere in the hierarchy 
eg.
```
const A = viewModelFactory({ id: new Field() });
A.fields
// contributeToClass called with A
class B extends A {}
B.fields
// contributeToClass called with B
class C extends B.augment({}) {}
C.fields
// contributeToClass called with C
```

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: explicit test case added

js-packages/@prestojs/viewmodel/src/fields/Field.ts
levic: Can you mention that in the header doc

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
levic: Can you give an example in header docs so that's clearer

js-packages/@prestojs/viewmodel/src/fields/Field.ts
davecoates: yep it's already there in the updates from yesterday

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
levic: Can you add a comment explaining what you just said, so that it's clearer why there is both a `delete()` and a `cleanupKey()`

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
levic: Can you expand that comment then, eg:
```
// TODO:Should this move up above this.pending.forEach?
```
to
```
// TODO:Should this move up above this.pending.forEach?
// (If any of the cb() calls trigger another batch then those changes will trigger
// an error because `pendingNoBatch` is set so that we don't lose any changes)
```
(Assuming what I just wrote is correct)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: There is in the latest changes: https://github.com/prestojs/prestojs/blob/feature/related-viewmodel-field/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts#L632

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: In latest commit I've removed this comment entirely with my changes to how listeners work

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
levic: The expanded explanation drops the mention of the fact that it doesn't include non-relation fields (prob want to update the examples to include one field, eg something along the lines of "note that the group's countryId field is included but not the country field because country is a RelatedViewModelField and wasn't specifically mentioned")

davecoates: I have a commit not yet pushed that says:

> We exclude [RelatedViewModelField](doc:RelatedViewModelField)'s to avoid circular dependencies. In the example above this means the `owner` field is excluded but the source field `ownerId` is still included.

(also note it's an internal presto function, not exported for public consumption)

PR #75, opened 2020-07-17
fix: add fieldClassName static prop to fields

### LINKED ISSUE #55: getWidgetForField relies on unminified class names
getWidetForField looks at class names rather than checking types with `instanceof` so can avoid importing all fields if they aren't required => smaller bundle sizes.

This doesn't work with minification however as the class names get mangled.

Options

1) Ditch checking names and just import all the fields

2) Add a required property or function on Field eg, `getClassName() { return 'FileField'; }`

davecoates:
This allows getWidgetForField to read this to determine what widget to return
for a field without needing to import the field class itself to use instanceof.
Resolves issues with using constructor.name which doesn't work with minified
builds.

Resolves #55


levic:
Looks fine.

PR #76, opened 2020-07-20
fix: remove FilterSet and FieldBinder

### LINKED ISSUE #47: FilterSet changes
Currently FilterSet is using the old `FieldBinder` implementation before I rewrote `ViewModel` as a factory function.

Most of the time a filter set has the same set of fields as a viewmodel but with some modifications (fields added / removed, different labels or help text).

I propose making filtersets a factory as well that you typically generate from a viewmodel

eg.

```
filterSetFactory({
   email: User.fields.email.clone({ helpText: '' }),
   dateJoined: new DateRangeField(),
})
```

I think the general case is that you'd use most fields the same so perhaps some kind of shortcut is in order

```
filterSetFactory(
  User.fields.clone({
    firstName: true,
    lastName: true,
    email: { helpText: '' },
  })
);
```

Implementation wise the differences between a ViewModel and FilterSet is that a FilterSet is never instantiated (it's juts a collection of fields) and has no cache associated with it.

davecoates:
Resolves #47


levic:
I was expecting changes to Form.Item -- did this already work with Form.Item and you were just removing the FilterSet way of doing it? (I was expecting this to be a lot more -- if this is really it then it looks fine)

davecoates:
Nope no changes to Form.Item, it really was just a different way of defining a collection of fields

PR #78, opened 2020-07-22
fix: throw on invalid field access

davecoates:
Only happens in dev. Gives much more useful feedback in various
circumstances eg. if you typo field name when using form:
  <Form.Field field={User.fields.firstNam} />
without this change you'll get an error from Field about field
or children being required.

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
+++ b/js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
@@ -890,6 +890,23 @@ export default function viewModelFactory<T extends FieldsMapping>(
             ];
             boundFields.set(modelClass as ViewModelConstructor<FinalFields>, f);
         }
+        if (isDev() && typeof Proxy != 'undefined') {
+            // In dev throw an error if invalid field is accessed
+            return [
+                new Proxy(f[0], {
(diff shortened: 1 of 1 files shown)


levic:
Looks fine.

Only question I have is whether there's anything we can do in prod (eg logging to Sentry). Not sure how you could do this in a way that isn't tied in to anything (`Console.warn()` maybe?)

davecoates:
> Looks fine.
> 
> Only question I have is whether there's anything we can do in prod (eg logging to Sentry). Not sure how you could do this in a way that isn't tied in to anything (`Console.warn()` maybe?)

Can just console.error instead of `throw` for prod. Originally I had the `dev` check because using `Proxy` which may not be supporte but as I'm checking that it exists anyway should be fine

davecoates:
Actually I'll just log an error in both cases so behaviour is same between dev and production

davecoates:
Actually I'm reconsidering whether this is a good idea. I originally added it after mistyping names (case mismatch) in dev and thought it would be useful to know... but it also means you can't test for field existence without triggering the error (eg. `if (model.fields.name) ...`). Internally presto does this a bit with ViewModelCache. I think on balance it's probably more annoying than useful

PR #79, opened 2020-07-23
Feature/endpoint middleware

### LINKED ISSUE #53: Endpoint middleware
In djrad the API layer by default handled:

1) If get an 401 response then the failed API call is put in a queue and once user logs in then it replays the request. For us to support this would need some concept of auth to handle. See https://gitlab.internal.alliancesoftware.com.au/alliance/alliance-redux-api#401-and-retries how it was handled (was relatively easy there due to using redux - could just respond to a specified auth action).

2) De-dupe get requests. If multiple requests were initiated to an API endpoint (GET method only) and all parameters were identical then only 1 request would be made - all calls to it would get the same promise.

It would be nice to support this in presto

davecoates:


js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -20,6 +20,15 @@ type ExecuteInitOptions = Omit<RequestInit, 'headers'> & {
     paginator?: PaginatorInterface | null;
 };
 
+/**
+ * Same as [fetch](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters) but
```
levic, line 24: Stating an assumption out aloud: I assume it's ok that we're relying on a fetch implementation and have no need to eg support XMLHttpRequest

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -96,10 +112,68 @@ export type ExecuteReturnVal<T> = {
...
+    execute: () => Promise<T>;
+    /**
+     * The original options passed to [execute](doc:Endpoint#method-execute)
+     */
+    executeOptions: EndpointExecuteOptions;
+};
+
+/**
+ * @param url The URL to call fetch with
+ * @param requestInit See [fetch parameteres](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters)
```
levic: parameteres => parameters

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -96,10 +112,68 @@ export type ExecuteReturnVal<T> = {
...
  */
 export type EndpointExecuteOptions = ExecuteInitOptions & UrlResolveOptions;
 
+type MiddlewareContext<T> = {
+    /**
+     * Function to re-execute the Endpoint. Accepts no arguments - it replays the request with identical arguments
+     * as the first time. This will go through the full middleware cycle again. This is useful for middleware
+     * that may replay a request after an initial failure, eg. if user isn't authenticated on initial attempt.
+     */
+    execute: () => Promise<T>;
```
levic, line 125: > it replays the request with identical arguments as the first time

that's not actually correct is it? Presumably it just uses executeOptions again, and if you've modified the options then it uses that

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -96,10 +112,68 @@ export type ExecuteReturnVal<T> = {
...
+    /**
+     * The original options passed to [execute](doc:Endpoint#method-execute)
+     */
+    executeOptions: EndpointExecuteOptions;
```
levic, line 129: How are you supposed to eg inject a header? Do you modify the EndpointExecuteOptions and then pass that through? If so, should mention that you should make a copy of it if you intend to modify it (you don't know what other middleware has done to the options)

levic: (or does the execute() method do this automatically for you?)

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -509,54 +643,107 @@ export default class Endpoint<ReturnT = any> {
             options = paginator.getRequestInit(restOptions);
         }
         const { urlArgs = {}, query, ...init } = options;
+        // Always make sure headers & method is set middleware implementations can assume it exists
```
levic: "set middleware" => "set so middleware"

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
```diff
@@ -0,0 +1,73 @@
+import { EndpointRequestInit, MiddlewareFunction } from './Endpoint';
+
+function defaultGetKey(url: string, requestInit: EndpointRequestInit): string {
+    return [
+        url,
+        requestInit.method,
+        [...requestInit.headers.entries()].map(([key, value]) => `${key}⁞${value}`).join('__'),
```
levic: See above; this will treat the order of headers as significant

(edge case: I don't think it's valid to have duplicate headers in HTTP is it?)

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
```diff
@@ -0,0 +1,73 @@
+import { EndpointRequestInit, MiddlewareFunction } from './Endpoint';
+
+function defaultGetKey(url: string, requestInit: EndpointRequestInit): string {
+    return [
+        url,
```
levic: If I have the same request with query params in a different order then it will be considered distinct. I can easily see the same query being constructed in 2 different places having params in a different order

I don't see any harm in also checking whether the query params match (order should only matter for the same item, ie
- `/foo?p=1&p=2` != `/foo?p=2&p=1`
- `/foo?a=1&b=2` = `/foo?b=2&a=1`

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
```diff
@@ -0,0 +1,73 @@
...
+     * Function to test whether a request is subject to de-duping. The default implementation only
+     * de-dupes requests with `method` set to `GET`.
+     */
+    test?: (url: string, requestInit: EndpointRequestInit) => boolean;
+    /**
+     * Function to get a key for a request. If the key matches another in flight request then it
+     * is considered a duplicate. The default implementation includes the URL, the request method and all request
+     * headers in the key.
+     */
+    getKey?: (url: string, requestInit: EndpointRequestInit) => string;
```
levic: Does this really have to return a string?

How many requests are actually going to be in-flight at once? Seems like it would be simpler and not have potential string escaping bugs if you allow an arbitrary data structure to be returned. Looking up a request match would mean iterating over each request key and doing a deep compare one at a time (so slower -- but if you only ever have at most 5 or 6 requests at once then still insignificant)

levic: (If performance became an issue, could add special-casing where the key is a string -- this would leave it up to the app if they wanted a simple implementation or a fast one)

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -509,54 +643,107 @@ export default class Endpoint<ReturnT = any> {
...
+                    }
+                    const nextMiddleware = middleware.shift();
+                    if (!nextMiddleware) {
+                        return runFetch(url, requestInit);
+                    }
+                    lastMiddleware = nextMiddleware;
+                    const r = nextMiddleware(url, requestInit, next, {
+                        execute: executeWithMiddleware,
+                        executeOptions: options,
+                    });
```
levic, line 735: - If there is some middleware that eg attempts to replay a request in the event of a 403 the then won't the first run have already removed everything from the `middleware` list, so the attempt at replaying the query will have no subsequent middleware?
- In the same example above, if middleware modifies the requestInit object then that will persist through both replays (eg if it adds a query param then the query param will have been added twice on the second replay)

It seems to me that before each invocation of a middleware next() needs to clone everything that's about to be passed into that middleware in case it's modified

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: Yeah I think so. The assumption is only for @prestojs/rest so if was really desired it would just be a new module

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: hmm yeah currently if they've been mutated you'll get the mutated options

I think it probably should do what I describe here though... perhaps need to take a copy when it executes

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: This is just exposing the original options for you here if needed. You don't pass the context through it's always passed by the middleware controller. The middleware function get passed `url`, `requestInit` and `context` - you make changes to `requestInit` (or `url`) before passing them to next function (you don't pass context).

Note that this is just the code documentation for what the context is an example of injecting headers is in the middleware docs

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
davecoates: yep good one, bad oversight by me

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
davecoates: the headers object can't have duplicates no

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
davecoates: good idea will solve issues with headers order

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: * no, at start of function we have `const middleware = [...this.middleware];` - it's copied. replaying it replays again with all middleware
* correct; as mentioned in your comment earlier mutations would change the context too so changing `next(url, requestInit);` to `next(url, deepClone(requestInit));` will resolve that to

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
```diff
@@ -0,0 +1,96 @@
+import { isDeepEqual } from '@prestojs/util';
+import qs from 'querystring';
+import { EndpointRequestInit, MiddlewareFunction } from './Endpoint';
+
+function defaultGetKey(
+    url: string,
+    requestInit: EndpointRequestInit
+): { url: string; method: string; headers: Record<string, string>; query: Record<string, string> } {
+    const [u, q] = url.split('?');
```
levic: This is not going to do the right thing with `"http://foo.bar/?a=4?b=6#?c=7"` or `"http://foo.bar/#?c=7"` or `"http://levi:passw?ord@foo.com/"` -- really need to use a lib function for the URL parsing

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
```diff
@@ -0,0 +1,96 @@
...
+    let query = {};
+    if (q) {
+        query = qs.parse(q);
+    }
+    return {
+        url: u,
+        query,
+        method: requestInit.method,
+        headers: [...requestInit.headers.entries()].reduce((acc, [key, value]) => {
+            acc[key] = value;
```
levic, line 20: According to my reading of query-string this will still work to differentiate
- `"http://foo.com/?a=1&a=2"`
- `"http://foo.com/?a=2&a=1"`
- `"http://foo.com/?a=2"`
- `"http://foo.com/?a=1"`

but can you add a test case for query strings with array values

js-packages/@prestojs/rest/src/__tests__/dedupeInFlightRequestsMiddleware.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    });
+    const p1 = action1.execute({ query: { param1: '1', param2: '2' } });
+    const p2 = action1.execute({ query: { param2: '2', param1: '1' } });
+    expect((await p1).result).toBe('hello world 0');
+    expect((await p2).result).toBe('hello world 0');
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    const p3 = action1.execute({ query: { param1: '1', param2: '2' } });
+    const p4 = action1.execute({ query: { param2: 'two', param1: '1' } });
+    expect((await p3).result).toBe('hello world 1');
+    expect((await p4).result).toBe('hello world 2');
```
levic, line 85: See comments elsewhere re test cases for:
- array values in a query string
- `?` chars in a URL that aren't part of the query string (username/password/fragment)
- multiple `?` chars in a query string

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -378,33 +452,91 @@ function isEqualPrepareKey(a: ExecuteInitOptions, b: ExecuteInitOptions): boolea
...
+ *   return next(url.toUpperCase(), requestInit)
+ * }
+ * ```
+ *
+ * This  middleware just transforms the response - converting it to uppercase.
+ *
+ * ```js
+ * function upperCaseResponseMiddleware(url, requestInit, next, context) {
+ *   const r = await next(url.toUpperCase(), requestInit)
+ *   return r.toUpperCase();
```
levic, line 480: Does that work? Isn't `r` a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response)?

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -378,33 +452,91 @@ function isEqualPrepareKey(a: ExecuteInitOptions, b: ExecuteInitOptions): boolea
...
-        paginatorClass: PaginatorInterfaceClass;
-    } = {
-        requestInit: {},
-        getPaginationState: defaultGetPaginationState,
-        paginatorClass: InferredPaginator,
-    };
+    /**
+     * This defines the default settings to use on an endpoint globally.
+     *
+     * All these options can be customised on individual Endpoint's.
```
levic: Endpoint's => Endpoints

js-packages/@prestojs/rest/src/Endpoint.ts
levic: Relying on the middleware to do the clone if it makes modifications seems like it's going to lead to bugs; people will forget to do it and it will cause bugs that are hard to track down (the documentation also doesn't mention that you'd need to do this that I can see).

Is there any reason you can't do a `deepClone()` on the options at the start of every `executeWithMiddleware()`?

levic (changes requested):

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: I think there's a misunderstanding - it does clone it (see where it kicks off it calls mergeRequestInit)

It's not expected that middleware would clone it - it's expected they mutate it but mutations only apply to current middleware execution chain.

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: yep. nope it's whatever your endpoint returns (eg. content type text/plain then it will be text)

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
davecoates: I've migrated usage of 'qs' library to 'query-string' which has good support for this and is smaller

(there was one rogue usage of 'querystring' as well - all use 'query-string')

js-packages/@prestojs/rest/src/dedupeInFlightRequestsMiddleware.ts
davecoates: done

js-packages/@prestojs/rest/src/__tests__/dedupeInFlightRequestsMiddleware.test.ts
davecoates: done

js-packages/@prestojs/rest/src/Endpoint.ts
levic: Can you mention this in the header for `execute()`.

I'm still unclear on how middleware would go about replaying a request though.

Consider the 403 example above.

I had thought that as the request came back up the middleware stack, the 403 interceptor would keep the original request unresolved, would trigger a login popup, log in again, replay the request from where we were at, and then return the request from that point on. If the middleware has already been removed then how can you replay the request from this point on?

What I thought (this uses promises so I realise the call stack is not really this, but conceptually I think it matches)

- execute()
  - executeWithMiddleware()
    - next()
      - customMiddlewareA
        - next()
          - 403Handler()
            - next()
              - customMiddlewareB() [modifies the request]
                - next()
                  - actual fetch request, fails with 403
                - resolve Promise with return from next()
            - 403 handler sees a 403 so does _not_ resolve promise, instead triggers a completely separate login cycle
            - 403 handler now tries to call next() again to replay the request from this point
              - but there's nothing left in the middleware array anymore
              - even if there was, customMiddlewareB will be called a second time and so will modify the request again. That's ok if it's idempotent but it may not be

levic: Edit: reading reviews in order sometimes helps!

With MiddlewareContext it would look like:

- execute()
  - executeWithMiddleware()
    - next()
      - customMiddlewareA
        - next()
          - 403Handler()
            - next()
              - customMiddlewareB() [modifies the request]
                - next()
                  - actual fetch request, fails with 403
                - resolve Promise with return from next()
            - 403 handler sees a 403 so does _not_ resolve promise, instead triggers a completely separate login cycle
            - 403 handler now calls MiddlewareContext.execute() again to trigger the request from the start

Is that correct?

davecoates: Not quite - the original promise wouldn't resolve (at least not how I would do it). The 403Handler instead returns a new promise that only resolves once the login happens. When it resolves it resolves to the promise returned from the replayed request after the successful login.

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -509,54 +643,107 @@ export default class Endpoint<ReturnT = any> {
             options = paginator.getRequestInit(restOptions);
         }
         const { urlArgs = {}, query, ...init } = options;
```
levic: The paginator is causing `options` to not be fully preserved for `execute()` later on; I could maybe see a use case if you have a limit on how many things you can fetch and want to chain say up to 10 pages of data at a time, but make it look to the caller as though you simply had a very large page size

levic:
lgtm

PR #80, opened 2020-08-03
Feature/file image upload

davecoates:
I'm going to add examples separately once I get it working with codesandbox

### DIFF
--- a/js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
+++ b/js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
@@ -31,12 +31,31 @@ export type UploadWidgetProps<FieldValue, T extends HTMLElement> = Omit<
     > & {
         input: UploadWidgetInputType;
         /**
-         * Number of files to allow. `null` means unlimited. Defaults to `1`.
+         * Number of files to allow. `null` means unlimited.
+         *
+         * If `limit` is `1` then after uploading a file the UI will change from an Add icon to an Edit icon and
+         * uploading a new file will replace the original.
+         *
+         * If `limit` is > `1` then once the limit has been reached the UI will disallow adding new files. To add a
+         * new file an existing one must be removed.
+         *
+         * If `limit` > `1` OR `multiple` is `true` then the value passed to `onChange` will be an array, otherwise
+         * it will be a `File` object or null. It is valid to set `limit` to `1` and `multiple` to `true`.
+         *
+         * Defaults to `1`.
          */
         limit: number | null;
         /**
-         * If true multiple files are accepted otherwise enforces a single value. When true `input.value` must be an
-         * array when set.
+         * Whether multiple values are accepted. In most cases you can set `limit` to `null` or a value greater than
+         * `1` instead of explicitly setting this. This can be set to `true` when `limit` is `1` in order to enforce
+         * the value being an array (eg. this could be useful if `limit` increases based on something else).
+         *
+         * If true `input.value` must be an array when set and `onChange` will be passed an array of
+         * [File](https://developer.mozilla.org/en-US/docs/Web/API/File) objects, otherwise it will be passed a single
+         * value or null.
+         *
+         * If `multiple` is `true` and `limit` is 1 then the only difference to `multiple={false}` is that the `value`
+         * will be an array.
          *
          * Defaults to `false` if `limit` is `1` otherwise `true`.
          */
@@ -194,8 +213,12 @@ export function useFileList(
 
 /**
  * File upload widget that wraps the antd [Upload](https://ant.design/components/upload/#API) component. Unlike
- * that component this one never uploads immediately - the raw `File` object will be passed to `onChange` and it's
- * expected the upload will happen externally (eg. on a form submit).
+ * that component this one never uploads immediately - the raw [File](https://developer.mozilla.org/en-US/docs/Web/API/File)
+ * object will be passed to `onChange` and it's expected the upload will happen externally (eg. on a form submit).
+ *
+ * If `multiple` is true then the value passed to `onChange` will be an array of `File` objects or an empty array when clearing the value.
+ *
+ * If `multiple` is false then the value passed to `onChange` will be a `File` object or `null` when clearing the value.
  *
  * For very custom requirements consider using the [useFileList](doc:useFileList) in a custom component.
  *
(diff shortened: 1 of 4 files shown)


js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
```diff
@@ -1,105 +1,317 @@
-import { PlusOutlined, UploadOutlined } from '@ant-design/icons';
+import { EditOutlined, InboxOutlined, PlusOutlined, UploadOutlined } from '@ant-design/icons';
```
levic, line 1: Can you confirm that tree shaking is working to remove unused icons?

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
```diff
@@ -1,105 +1,317 @@
...
+         * Number of files to allow. `null` means unlimited. Defaults to `1`.
+         */
+        limit: number | null;
+        /**
+         * If true multiple files are accepted otherwise enforces a single value. When true `input.value` must be an
+         * array when set.
+         *
+         * Defaults to `false` if `limit` is `1` otherwise `true`.
+         */
+        multiple: boolean;
```
levic, line 62: This seems unnecessary since it can be calculated from `limit > 1`

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
```diff
@@ -1,105 +1,317 @@
...
+                name,
+                size: blob.size,
+                type: blob.type,
+                thumbUrl: reader.result as string,
+                originFileObj: blob,
+            })
+        );
+        reader.addEventListener('error', error => reject(error));
+        try {
+            reader.readAsDataURL(blob);
```
levic, line 92: Are we base64 encoding uploads? That's going to blow out upload sizes by 50% (cf readAsBinaryString).

(Or are you assuming it's always going to be embedded in the json?)

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
```diff
@@ -1,105 +1,317 @@
...
+    props: UploadWidgetProps<File, HTMLInputElement>,
+    ref: React.RefObject<Upload>
+): React.ReactElement {
+    let {
+        input,
+        children,
+        beforeUpload,
+        listType = 'text',
+        limit = 1,
+        multiple = limit !== 1,
```
levic, line 245: Need to comment the interactions between limit and multiple, esp

- what does it mean to have `limit == 1` and `multiple=true`?
- What changes the way the UI is represented? (seems to be `limit == 1`)
- If you have `multiple` set then `file` will always be an array? If `limit ==1 && multiple == true` then will `file` be a single value or will it be an array?
- if `limit == 1` does that mean there's always going to be 1 file or does it mean "0 or 1" files?

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
```diff
@@ -1,105 +1,317 @@
...
+            onChange(file);
+        }
+        return false;
+    };
+    const handleRemove = (f: FileWidgetUploadFile): void => {
+        if (!multiple) {
+            onChange(null);
+        } else if (value) {
+            onChange((value as (string | File)[]).filter(key => key !== f.key));
+        }
```
levic, line 295: Can you swap those if clauses: this is different from the other places where `multiple` is checked (the `multiple` part always comes first)

levic (changes requested):

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
davecoates: based on other discussions I assume you are ok to keep this? (eg. to allow limit === 1, multiple === true)

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
davecoates: As discussed only used in preview for images

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
davecoates: yep

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
```diff
@@ -31,12 +31,31 @@ export type UploadWidgetProps<FieldValue, T extends HTMLElement> = Omit<
...
-         * Number of files to allow. `null` means unlimited. Defaults to `1`.
+         * Number of files to allow. `null` means unlimited.
+         *
+         * If `limit` is `1` then after uploading a file the UI will change from an Add icon to an Edit icon and
+         * uploading a new file will replace the original.
+         *
+         * If `limit` is > `1` then once the limit has been reached the UI will disallow adding new files. To add a
+         * new file an existing one must be removed.
+         *
+         * If `limit` > `1` OR `multiple` is `true` then the value passed to `onChange` will be an array, otherwise
```
levic: limit > 1 and multiple==false is ~should really be~ is an error, in which case that condition is just:

> If `multiple` is `true` then ...

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
levic: If multiple is true then value will always be true (will always be a non-empty array otherwise you can't have clicked on the handleRemove button) in which case the same if pattern as the others applies:

```
        if (multiple) {
            onChange((value as (string | File)[]).filter(key => key !== f.key));
        } else {
            onChange(null);
        }
```

(ie unless I've read this wrong the check for value is unnecessary)

davecoates: yep good point, fixed

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
davecoates: updated

levic:
lgtm

PR #81, opened 2020-08-10
Feature/blank required

davecoates:
See https://gitlab.internal.alliancesoftware.com.au/alliance/presto/-/issues/2

Was originally done by fang in https://github.com/prestojs/prestojs/pull/24/files, just pulled out the blank/blankAsNull stuff

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/fields/Field.ts
+++ b/js-packages/@prestojs/viewmodel/src/fields/Field.ts
@@ -7,9 +7,16 @@ import { AsyncChoicesInterface } from './AsyncChoices';
  */
 export interface FieldProps<T> {
     /**
-     * True if field is required when creating or updating a model
+     * Is this field required when saving a record?
      */
-    required?: boolean;
+    blank?: boolean;
+    /**
+     * If true an empty string value should be converted to a null value
+     *
+     * This is useful when backend considers an empty string different to `null` and specifically
+     * needs to handle empty values as `null`.
+     */
+    blankAsNull?: boolean;
     /**
      * Label for this field. If not specified will be generated from the name.
      */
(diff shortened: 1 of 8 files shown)


js-packages/@prestojs/viewmodel/src/fields/Field.ts
```diff
@@ -7,9 +7,16 @@ import { AsyncChoicesInterface } from './AsyncChoices';
  */
 export interface FieldProps<T> {
     /**
-     * True if field is required when creating or updating a model
+     * Is this field required when saving a record?
      */
-    required?: boolean;
+    blank?: boolean;
```
levic, line 12: Comment doesn't match the name, should be more explicit

"Is this field allowed to be assigned a blank (null, undefined, '') value?"

js-packages/@prestojs/viewmodel/src/fields/Field.ts
```diff
@@ -7,9 +7,16 @@ import { AsyncChoicesInterface } from './AsyncChoices';
...
+    /**
+     * If true an empty string value should be converted to a null value
+     *
+     * This is useful when backend considers an empty string different to `null` and specifically
+     * needs to handle empty values as `null`.
+     */
+    blankAsNull?: boolean;
```
levic, line 20: This is on the base `Field.ts` class

What does it mean for an `IntegerField`? Or eg an IP Address field?

I would have thought that this really only has meaning for string-based fields (textfield/charfield) where `""` is a potentially valid value. Should this be moved down into CharField/TextField? (or possibly an intermediate AbstractStringField class?)

levic (changes requested):

js-packages/@prestojs/viewmodel/src/fields/Field.ts
davecoates: I'm not sure it's worth it. Frontend values are often strings even if they aren't stored like that in the database and experience tells me pushing it all to the backend is more pain than it's worth (eg. with DRF you'll have to customise fields just to change "" to `null` for a decimal value to parse).

So your IP address field example - backend likely expects a string value for the IP address. Does it handle empty string for no value or expect null? Depends on the backend... so supporting this generic conversion on the base field makes sense to me.

If it was a new base class how would `DecimalField` implement it for example (it already extends the base `NumberField`)? It would have to duplicate the functionality (we can also solve this by converting "" to `null` in DecimalField specifically but do we want to?). I'm inclined to leave it in the base for this reason.

PR #87, opened 2020-09-11
fix: endpoint method defaults

davecoates:
Fix bug that meant if you provided a method in the constructor it would
be replaced with 'GET' when execute was called (execute only respected
method if it was passed in the execute call)


js-packages/@prestojs/rest/src/__tests__/Endpoint.test.ts
```diff
@@ -125,6 +125,25 @@ test('should support transformation function', async () => {
     expect((await action2.prepare().execute()).result).toEqual({ b: 'a', d: 'c' });
 });
 
+test('should support merging endpoint options and ezecute options', async () => {
```
levic, line 128: ezecute => execute

levic:
lgtm

PR #96, opened 2020-09-24
ListField & ManyRelatedViewModelField

davecoates:
As part of doing this I've found various edge cases where the caching fails to do what you would expect in regards to listeners.

I've added failing test cases for these that are currently skipped. I spent many hours trying to resolve it but it's going to take a fair bit to get there I think... probably another rewrite of the internals of caching as it's really difficult to follow.

These issues exist with or without this branch however so think we may as well proceed with this as is. This extends the existing behaviour to support RelatedViewModelField's that have multiple values instead of a single value.

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
+++ b/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
@@ -1,5 +1,6 @@
+import { isEqual as isShallowEqual } from '@prestojs/util';
 import pick from 'lodash/pick';
-import RelatedViewModelField from './fields/RelatedViewModelField';
+import { BaseRelatedViewModelField } from './fields/RelatedViewModelField';
 import { isDev } from './util';
 import {
     expandRelationFieldPaths,
@@ -167,7 +168,11 @@ function getFieldNameCacheKey(
         } else {
             // getField guarantees that the field will be a RelatedViewModelField so we
             // don't need to check it here
-            const relatedField = viewModel.getField(path.slice(0, -1)) as RelatedViewModelField;
+            const relatedField = viewModel.getField(path.slice(0, -1)) as BaseRelatedViewModelField<
+                any,
+                any,
+                any
+            >;
             if (!relatedField.to.pkFieldNames.includes(path[path.length - 1])) {
                 flatFieldNames.add(path.join('.'));
             }
@@ -279,7 +284,7 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
                 } else {
                     relations[fieldName[0]].push(fieldName.slice(1));
                 }
-            } else if (viewModel.fields[fieldName] instanceof RelatedViewModelField) {
+            } else if (viewModel.fields[fieldName] instanceof BaseRelatedViewModelField) {
                 relations[fieldName] = relations[fieldName] || [];
                 nonRelationFieldNames.push(viewModel.fields[fieldName].sourceFieldName);
             } else {
@@ -294,7 +299,9 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
                 // If field names for relation haven't been specified default to all non-relation fields
                 relationFieldNames.push(
                     ...relation.to.fieldNames.filter(fieldName => {
-                        return !(relation.to.fields[fieldName] instanceof RelatedViewModelField);
+                        return !(
+                            relation.to.fields[fieldName] instanceof BaseRelatedViewModelField
+                        );
                     })
                 );
             }
@@ -437,7 +444,7 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
                             if (record) {
                                 if (!after) {
                                     this.cleanupKey(fieldsKey);
-                                } else if (!after.isEqual(record[fieldName])) {
+                                } else if (!isShallowEqual(after, record[fieldName])) {
                                     const newRecord = new this.viewModel({
                                         ...record._data,
                                         [fieldName]: after,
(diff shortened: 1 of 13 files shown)


js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -167,7 +168,11 @@ function getFieldNameCacheKey(
         } else {
             // getField guarantees that the field will be a RelatedViewModelField so we
             // don't need to check it here
-            const relatedField = viewModel.getField(path.slice(0, -1)) as RelatedViewModelField;
+            const relatedField = viewModel.getField(path.slice(0, -1)) as BaseRelatedViewModelField<
+                any,
+                any,
+                any
+            >;
```
levic, line 175: Does `any` here really mean any, or does it mean "any, but as long as it satisfies the 'extends' constraints in the class definition"?

js-packages/@prestojs/viewmodel/src/fields/Field.ts
```diff
@@ -4,7 +4,7 @@ import { AsyncChoicesInterface } from './AsyncChoices';
 /**
  * @expand-properties
  */
-export interface FieldProps<T> {
+export interface FieldProps<T, SingleType = T> {
```
levic: Can you add a comment for what T and SingleType are -- it's not clear at first glance

(I assume T is the value in the field which in the case of multivalue fields would normally be an array of SingleType?)

js-packages/@prestojs/viewmodel/src/fields/Field.ts
```diff
@@ -67,7 +67,7 @@ class UnboundFieldError<T, K> extends Error {
  * @extract-docs
  * @menu-group Fields
  */
-export default class Field<T, ParsableType extends any = T> {
+export default class Field<T, ParsableType extends any = T, SingleType = T> {
```
levic: Is it documented anywhere what ParsableType is?

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
```diff
@@ -1,11 +1,31 @@
...
+type RelatedViewModelParsableType<T extends ViewModelConstructor<any>> =
+    | FieldDataMappingRaw<T['fields']>
+    | FieldDataMappingRaw<T['fields']>[];
 
 /**
  * @expand-properties
  */
-type RelatedViewModelFieldProps<T extends ViewModelConstructor<any>> = FieldProps<
-    ViewModelInterface<T['fields'], any, T['__pkFieldType'], T['__pkType']>
+type RelatedViewModelFieldProps<T extends ViewModelConstructor<any>, FieldValueType> = FieldProps<
```
levic: As a general comment, am wondering about being more explicit with types. `T` is helpful because you immediately know that it's a generic type, but it's not always obvious what the type represents.

Does ending them with T (to distinguish from regular types) make sense?

eg in this case it would be something like
`type RelatedViewModelFieldProps<TargetViewModelT extends ViewModelConstructor<any>, FieldValueT>`

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
```diff
@@ -20,15 +40,148 @@ type RelatedViewModelFieldProps<T extends ViewModelConstructor<any>> = FieldProp
...
+            this._resolvedTo = to;
+        } else {
+            this._loadTo = to;
+        }
+        this.sourceFieldName = sourceFieldName;
+        this._isResolvingDeps = false;
+    }
+
+    /**
+     * @private
```
levic, line 94: Should document what this function does -- as with comment on other review, it will make sense to devs familiar with django internals but not necessarily other frameworks/languages

levic (commented):
Happy for this to be committed without further review (but can you @ me so I can see the answers to my questions)

levic:
I don't see any issue for refactoring the caching code; can you add an item to the board for this.

Be reasonably specific about what you think needs to be updated -- the test cases are pretty comprehensive but if you're not around I think someone looking at it will have no idea where to start at all.

js-packages/@prestojs/viewmodel/src/fields/Field.ts
davecoates: Yep that's right

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
davecoates: Was trying to find some existing conventions in community, best I came across was this https://docs.microsoft.com/en-gb/archive/blogs/kcwalina/design-guidelines-update-names-of-generic-type-parameters which was referenced in a discussion about typescript. Seems as good as any 🤷‍♂️

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
davecoates: It's documented on the base class

js-packages/@prestojs/viewmodel/src/fields/Field.ts
davecoates: No generally haven't documented generic types, will add something

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: pretty sure it results in all constraints being ignored. i don't think it really matters here though as don't know anything specific... i guess could specify the constraints again 🤷‍♂️

js-packages/@prestojs/viewmodel/src/fields/RelatedViewModelField.ts
levic: Using T as a prefix seems a very Microsoft thing to do (eg everything in COM is prefixed with I for Interface, ), but suffixes seem more common in JS-land (but my exposure to other typescript codebases is not very strong so I may be wrong)

PR #97, opened 2020-09-30
Feature/endpoint middleware refactor

### LINKED ISSUE #84: ViewModelEndpoint delete
Currently no way to handle a delete in a ViewModelEndpoint

Don't see an easy way to achieve this with current hooks - response not available in transformResponseBody. May need something else?

davecoates:
Resolves #84


django-testbed/frontend/src/models/User.js
```diff
@@ -13,21 +13,30 @@ import namedUrls from '../namedUrls';
...
             });
         },
     }),
-    create: new ViewModelEndpoint(namedUrls.get('users-list'), () => User, {
+    create: new Endpoint(namedUrls.get('users-list'), {
         method: 'post',
+        middleware,
+    }),
+    delete: new Endpoint(namedUrls.get('users-detail'), {
+        method: 'DELETE',
```
levic, line 38: Minor consistency: methods should all be the same case (presumably uppercase due to your issue last week)

django-testbed/frontend/src/models/User.js
```diff
@@ -13,21 +13,30 @@ import namedUrls from '../namedUrls';
...
     }),
-    list: new PaginatedViewModelEndpoint(namedUrls.get('users-list'), () => User, {
+    list: new Endpoint(namedUrls.get('users-list'), {
+        middleware: [...middleware, paginationMiddleware()],
         resolveUrl(urlPattern, urlArgs, query) {
             return urlPattern.resolve(urlArgs, {
                 // default to pageNumber pagination if not specified
                 query: { ...query, paginationType: query.paginationType || 'pageNumber' },
             });
         },
```
levic, line 31: Can `resolveUrl` be a part of the middleware? This seems like something you're going to duplicate every time you use paginationMiddleware

django-testbed/frontend/src/views/UserListView.js
```diff
@@ -21,6 +21,11 @@ export default function UserListView() {
     const paginationStatePair = useNaviUrlQueryState({}, { prefix: 'p_' });
     const [selectedId, selectId] = useState();
     const [showCreate, setShowCreate] = useState(false);
+    const { run: runDelete } = useAsync(id =>
+        User.endpoints.delete.execute({
+            urlArgs: { id },
+        })
+    );
```
levic, line 28: Is this a pattern that should be part of each endpoint? Seems like something that's going to be duplicated every time you want to use an endpoint

Something along these lines:

```js
const runDelete = User.endpoints.useDelete();
const runDelete = User.endpoints.delete.useExecute();
const { runDelete } = User.endpoints.useExecute(); // all of the endpoints in one go
```

js-packages/@prestojs/viewmodel/src/index.ts
```diff
@@ -70,4 +73,7 @@ export type {
     AsyncChoicesInterface,
     AsyncChoicesOptions,
     Choice,
+    PrimaryKey,
+    CompoundPrimaryKey,
+    SinglePrimaryKey,
```
levic, line 78: Are the changes in this file supposed to be from this branch? (Is this just a branching an artifact?)

js-packages/@prestojs/rest/src/paginationMiddleware.ts
```diff
@@ -0,0 +1,117 @@
...
+                    urlArgs: urlConfig.args,
+                    query: urlConfig.query,
+                });
+                Object.assign(urlConfig, urlConfigChanges);
+                requestInit = mergeRequestInit(requestInit, { headers }) as EndpointRequestInit;
+            }
+            const response = await next(urlConfig, requestInit);
+            if (paginator) {
+                if (response !== context.decodedBody) {
+                    throw new Error('paginatorMiddleware must be the first to handle the response');
```
levic, line 108: Can you add an // explanation with some more details as to why this is the case

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * @expand-properties
+ */
+type ViewModelCachingOptions<T> = {
+    /**
+     * A function to return the ID to used when a DELETE occurs. This id is used to remove the item with that ID
+     * from the cache.
+     *
+     * Defaults to returning the url argument `id` (eg. from a UrlPattern like `/users/:id`).
```
levic, line 41: For reference, I was curious whether this default would be enough and (at least in the official docs) rails, yii, symfony & .NET all use "id" as the identifier

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
```diff
@@ -0,0 +1,197 @@
...
+        contributeToClass(endpoint: Endpoint): void {
+            // When using the default implementation we can check things are setup correctly on initialisation and
+            // throw an error. For custom implementations we have to wait until the method is called to do the check.
+            if (getDeleteId === defaultGetDeleteId && endpoint.requestInit.method === 'DELETE') {
+                if (!endpoint.urlPattern.requiredArgNames.includes('id')) {
+                    throw new Error(
+                        `When using 'viewModelCachingMiddleware' on a DELETE endpoint it is expected the UrlPattern includes an 'id' parameter. Known parameters are: ${endpoint.urlPattern.validArgNames.join(
+                            ', '
+                        )}. You can pass 'getDeleteId' to override this behavior.`
+                    );
```
levic: It seems to break encapsulation for viewModelCachingMiddleware to know about the implementation of defaultGetDeleteId -- can this check be somehow made part of the `defaultGetDeleteId` interface?

not exactly sure how to do this without complicating the `getDeleteId` interface though -- to keep it simple you still want it to be a function
- could add a function to the function (eg `getDeletedId.validateEndpoint()`). It a bit out of the ordinary
- Add a `validateEndpoint` to `ViewModelCachingOptions` to -- but then if you override getDeleteId you'd also need to override `Endpoint`, or have the default implementation check `getDeleteId` and do nothing if it's not the default. Is getting a bit messy

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
```diff
@@ -0,0 +1,197 @@
...
+            const response = await next(urlConfig, requestInit);
+            if (context.requestInit.method?.toUpperCase() === 'DELETE') {
+                const _viewModelMapping = await resolveViewModelMapping();
+                if (!isViewModelClass(_viewModelMapping)) {
+                    throw new Error(
+                        'When handling DELETE the view model mapping must be a single ViewModelClass'
+                    );
+                }
+                const id = getDeleteId(context);
+                if (id == null) {
```
levic: Could be an empty string or undefined?

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -63,12 +54,22 @@ export type EndpointOptions = ExecuteInitOptions & {
      */
     resolveUrl?: (urlPattern: UrlPattern, urlArgs?: Record<string, any>, query?: Query) => string;
     /**
-     * Middleware to apply for this endpoint. Defaults to [Endpoint.defaultConfig.middleware](http://localhost:3000/docs/rest/Endpoint#static-var-defaultConfig) - global
+     * Middleware to apply for this endpoint. Defaults to [Endpoint.defaultConfig.middleware](doc:Endpoint#static-var-defaultConfig) - global
```
levic: That's no longer correct: it default to empty, and getMiddleware() appends the global middleware (maybe say something along the lines of "by default getMiddleware() concatenates this with the global middleware")

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -83,7 +84,7 @@ export type ExecuteReturnVal<T> = {
     /**
      * The url that the endpoint was called with
      */
-    url: string;
+    url?: string;
```
levic: How can you make a HTTP request without a url?

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -63,12 +54,22 @@ export type EndpointOptions = ExecuteInitOptions & {
...
+     * Middleware to apply for this endpoint. Defaults to [Endpoint.defaultConfig.middleware](doc:Endpoint#static-var-defaultConfig) - global
      * middleware options can be set there.
      *
      * See [middleware](#Middleware) for more details
      */
-    middleware?: MiddlewareFunction<any>[];
+    middleware?: Middleware<ReturnT>[];
+    /**
+     * Get the final middleware to apply for this endpoint. This combines the global middleware and the middleware
+     * specific to this endpoint. Defaults to [Endpoint.defaultConfig.getMiddleware](doc:Endpoint#static-var-defaultConfig).
```
levic: Specify the order (middleware + global middleware by default)

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -127,6 +127,38 @@ type MiddlewareContext<T> = {
...
      */
     executeOptions: EndpointExecuteOptions;
+    /**
+     * The endpoint the middleware is executing within.
+     */
+    endpoint: Endpoint;
+    /**
+     * The resolved URL.
+     *
+     * Only available in the response part of the middleware.
```
levic, line 145: Is it worthwhile making 2 types, MiddlewareContextResponse which extends MiddlewareContextRequest so that TS can catch the errors that you are doing in the getters? (applies to the other post-request items too)

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -427,9 +470,9 @@ function isEqualPrepareKey(a: ExecuteInitOptions, b: ExecuteInitOptions): boolea
  *
  * ## Pagination
  *
- * Pagination for an endpoint is handled by a `Paginator` class returned from the `getPaginatorClass` method. The base
- * `Endpoint` class defaults to using no pagination. [PaginatedEndpoint](doc:PaginatedEndpoint) is provided with a
- * default implementation that chooses the paginator based on the shape of the response (eg. if the response looks like
+ * Pagination for an endpoint is handled by [paginationMiddleware](doc:paginationMiddleware). This middleware
+ * will add a `getPaginatorClass` method to the `Endpoint` which makes it compatible with [usePaginator](doc:usePaginator).
```
levic, line 491: > will add a `getPaginatorClass` method to the `Endpoint`

only if not already present?

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -340,13 +381,15 @@ function isEqualPrepareKey(a: ExecuteInitOptions, b: ExecuteInitOptions): boolea
...
- * decoding the data based on content type and is suitable for endpoints that return JSON or text with the appropriate
- * content types. If you just wish to do something with the decoded data (eg. the JSON data) use `transformResponseBody`.
- *
- * In addition you can pass all options accepted by `fetch` and these will be used as defaults to any call to `execute`
- * or `prepare`.
+ * * Accepts a [UrlPattern](doc:UrlPattern) to define the URL used. Any arguments & query parameters can be passed at execution time
+ * * Accepts a `decodeBody` function that decodes the `Response` body as returned from [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). The
+ *   default `decodeBody` will interpret the response based on the content type
+ *   * If type includes 'json' (eg. application/json) returns decoded json
+ *   * If type includes 'text (eg. text/plain, text/html) returns text
```
levic, line 405: I didn't pick this up before but this description does match the implementation in `decodeBody`.

Those rules are going to catch things they really shouldn't, eg `application/ubjson` (binary json)

[At the end of this section](https://mimesniff.spec.whatwg.org/#mime-type-groups) is a good heuristic (the +json syntax is used by at least github). Also content type may contain character encoding (eg `text/plain; charset=UTF-8`), does the browser deal with that for you and makes sure it's always utf8? Does it appear in the content type or does the browser strip the character encoding off?

I think for text I'd do a "text/" prefix. This will still treat "text/xml" as text but we're not parsing xml anyway.

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -511,6 +554,38 @@ function isEqualPrepareKey(a: ExecuteInitOptions, b: ExecuteInitOptions): boolea
  * new Endpoint('/users/', { middleware: [csrfTokenMiddleware] })
  * ```
  *
+ * When middleware is passed to the `Endpoint` it is _appended_ by to the default
```
levic: "by to" => to"

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -511,6 +554,38 @@ function isEqualPrepareKey(a: ExecuteInitOptions, b: ExecuteInitOptions): boolea
...
+ *   requestDefaultsMiddleware,
+ *   ...Endpoint.defaultConfig.middleware,
+ *   ...middleware,
+ * ]
+ * ```
+ *
+ * The [requestDefaultsMiddleware](doc:requestDefaultsMiddleware) is included here
+ * rather than in `Endpoint.defaultConfig.middleware` as it's strongly recommended
+ * to have this on and would be easy to inadvertently remove it from the global
+ * middleware.
```
levic: I think that make it worse

You now have to remember to include `requestDefaultsMiddleware` every time you override `getMiddleware()` instead of having to remember to put it once in the global middleware.

Using django as an example: there's a bunch of middleware that's there in every project, you're always starting from a default list instead of creating it from scratch anyway

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -527,45 +602,39 @@ export default class Endpoint<ReturnT = any> {
...
-     * Get the Paginator class to use for this endpoint.
-     *
-     * `Endpoint` defaults to no paginator. See [PaginatedEndpoint](doc:PaginatedEndpoint)
-     * for class that provides default paginator.
-     */
-    getPaginatorClass(): PaginatorInterfaceClass<any> | null {
-        return null;
+        this.middleware = getMiddleware(middleware, this);
+        this.middleware.forEach(m => {
+            if (typeof m !== 'function' && m.contributeToClass) {
```
levic: contributeToClass seems even less apt now: it's not modifying a class, it's modifying an endpoint instance

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -511,6 +554,38 @@ function isEqualPrepareKey(a: ExecuteInitOptions, b: ExecuteInitOptions): boolea
...
+ * middleware.
+ *
+ * You can change the default implementation on [Endpoint.defaultConfig.getMiddleware](doc:Endpoint#static-var-defaultConfig)
+ *
+ * Middleware can also be defined as an object with any of the following properties:
+ *
+ * * `contributeToClass` - Called when the `Endpoint` is initialised and allows the middleware to modify the endpoint
+ *   class or otherwise do some kind of initialisation.
+ * * `prepare` - A function that is called in `Endpoint.prepare` to modify the options used. Specifically this allows middleware
+ *   to apply it's changes to the options used (eg. change URL etc) such that `Endpoint` correctly caches the call.
```
levic: it's => its

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -760,7 +837,10 @@ export default class Endpoint<ReturnT = any> {
 // Initialisation is not done inline in class as doc extractor doesn't handle object literals well
 Endpoint.defaultConfig = {
     requestInit: {},
-    getPaginationState: defaultGetPaginationState,
-    paginatorClass: InferredPaginator,
     middleware: [],
+    getMiddleware: <T>(middleware: Middleware<T>[]): Middleware<T>[] => [
+        requestDefaultsMiddleware,
```
levic: See comment earlier re this going into global defaults

django-testbed/frontend/src/models/User.js
davecoates: it's only used here to add some defaults to `query` if not otherwise set (usually would do this at the view level anyway, don't remember why was done here)

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -669,49 +728,54 @@ export default class Endpoint<ReturnT = any> {
...
+                    .then(async res => (returnVal.decodedBody = await this.decodeBody(res)));
             const executeWithMiddleware = (): Promise<ReturnT> => {
                 const middleware = [...this.middleware];
                 let lastMiddleware;
+                const middlewareContext: MiddlewareContext<ReturnT> = {
+                    execute: executeWithMiddleware,
+                    executeOptions: options,
+                    requestInit,
+                    endpoint: this,
+                    get url(): string {
```
levic: See comments earlier: if you have a pre/post fetch middleware contexts then typescript should be able to catch these errors

django-testbed/frontend/src/views/UserListView.js
davecoates: something to consider, more likely `useEndpoint(User.endpoints.delete)` (which we have in the project itself currently)

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: no but it's initially not set until the url is resolved

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: no because you only have one type in the middleware, eg.

```
function myMiddleware(urlConfig, requestInit, next, context) {
   // currently url not available
  await next(urlConfig, requestInit);
  // now it is but `context` is still the same
}
```

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: Not specific to this branch, opened #100

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: well it currently always does it

guess it can check and throw an error

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: as above not possible

and still really want to catch them at runtime if not using typescript

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
davecoates: it does but it's about providing better dev experience over purity

i kinda don't think it's worth making it more complex - if you are overriding getDeleteId you are probably doing it for a specific endpoint and wouldn't be worth writing the code to do that check

i guess if your project used a different convention altogether it might be wroth it

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
davecoates: `==  null` catches undefined

was avoiding `(!id)` as `0` would be valid. will change to `id == null || id === ''`

js-packages/@prestojs/viewmodel/src/index.ts
davecoates: needed to be able to import these things in a different package

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: ok will move it

not convinced having this middleware was the best idea given other middleware relies on it but i guess it gives ultimate flexibility

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: ok, I will change it to `init`

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: i've changed how this is used so can be typed as the final value rather than needing it to describe the intermediate values which are internal

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: i've added to the docs the recommendation to add to the global middleware rather than replacing it

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
davecoates: ok this was easy enough just to add a `validateEndpoint` function so i've moved the check in there

django-testbed/frontend/src/models/User.js
levic: I don't quite understand -- this looks like it's just adding some default params for pagination purposes.

If you have to do that on the endpoint then won't you have to do that on every endpoint that uses pagination? And if that's the case then shouldn't it be pagination that takes care of it?

js-packages/@prestojs/rest/src/Endpoint.ts
levic: Yes, you would end up with 2 variables:

```js
function myMiddleware(next, urlConfig, requestInit, requestContext) {
   // currently url not available
  responseContext = await next(urlConfig, requestInit, requestContext);
  // now it is in `responseContext`
}
```

(have reordered the order of arguments to make it a little clearer)

davecoates: I don't particularly like `next` returning context instead of response. The middleware returns the response (and it should... too cumbersome and error prone to expect it to also return context) so seems confusing for `next` to return context instead.

I'm inclined to leave it with runtime checking given changing it would really only give you slightly better types at the cost of a more complicated interface

django-testbed/frontend/src/models/User.js
davecoates: In the testbed you choose pagination type for test purposes - page number, cursor, offset/limit - this is just setting which type to use by default. Not relevant to general case.

js-packages/@prestojs/rest/src/Endpoint.ts
davecoates: It is nice having the URL and response etc only available after the call to next but not sure what the return should be from `next`. It needs to be possible just to return direct so we'd have to support that use case, eg.

```js
function requestOnlyMiddleware(urlConfig, requestInit, next, context) {
   // do something to requestInit
   return next(urlConfig, requestInit);
}
```

If next was to return the relevant context for response that would give access to `url`, `response` and `decodedBody`. But most of the time you just want to do something to the transformed decodedBody from the previous middleware... so how should that be accessed? Another key, eg `result`?

```js
function middleware(urlConfig, requestInit, next) {
  const { decodedBody, url, response, result } = await next(urlConfig, requestInit);

  return transformAndCache(result);
}
```

or array return

```js
function middleware(urlConfig, requestInit, next) {
  const [result, { decodedBody, url, response }] = await next(urlConfig, requestInit);
  // Or if you don't need any of the response context:
  //   const [result] = await next(urlConfig, requestInit);

  return transformAndCache(result);
}
```

Any thoughts?

Do we want to change the order of arguments to have `next` first?

django-testbed/frontend/src/models/User.js
levic: Ah I misunderstood, I thought this would always be required

js-packages/@prestojs/rest/src/Endpoint.ts
levic: Yes please to changing order of arguments.

To clarify what you're saying: the return value from `next` will be the result + the fields that were previously only available to MiddlewareContext after the request?

For fields that were available prior to that you'd just use the same MiddlewareContext you had before -- there's nothing the other middleware could have changed that we'd care about? (eg executeOptions?)

re the syntax, I prefer the first one. From the perspective of a user I don't see any difference between `result` and `url` -- they're still all just the output of the next() call.

davecoates: The difference is you have to return result whereas url is there in case you need it

levic:
lgtm

PR #102, opened 2020-10-07
fix: getAll works with nested fields and *

### LINKED ISSUE #82: RelatedViewModelField improvements/fixes
- [x] Export from index.ts
- [x] Support lists of values. Probably requires support for lists in general for the source field to work
- [ ] `ViewModelCache.getAll` doesn't seem to support getting related fields, at least not properly eg. `cache.getAll([['operator', 'id']])` works but `cache.getAll([['operator', 'abbreviatedName']])` doesn't despite the field existing (and it works if you call `.get(id, ...)`

davecoates:
Also fix types on getList so accepts '*'

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
+++ b/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
@@ -1147,6 +1147,34 @@ export default class ViewModelCache<
     }
 
     /**
+     * Resolves '*' to all fields and validates passed fields are all valid
+     */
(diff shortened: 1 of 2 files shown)


levic:
unless I'm misunderstanding, this is just exposing functionality that's actually already there into other places?

lgtm

davecoates:
> unless I'm misunderstanding, this is just exposing functionality that's actually already there into other places?
> 
> lgtm

yep

PR #103, opened 2020-10-07
feature: useUrlQueryState work with History by default

### LINKED ISSUE #90: useUrlQueryState default implementation
The default implementation can almost be used as is with the default History api with one shortcoming - nothing causes a rerender

Currently you can workaround this:

```
    const [, forceRender] = useReducer(s => !s, false);
    const replaceUrl = (url: string): void => {
        window.history.pushState(null, '', url);
        forceRender();
    };
    const [filter, setFilter] = useUrlQueryState(
        {},
        {
            replaceUrl,
        }
    );
```

I propose we make the forceRender part of our default `replaceUrl` implementation

davecoates:
Resolves #90


levic:
lgtm

am not entirely sure how you handle the type info being different if you are using a different location handler api but that can be handled later

ie this definition may be different:
```js
type Options = {

    location?: {
        search: string;
        pathname: string;
    };
```

davecoates:
It doesn't matter - you just have to pass an object with those properties. So you do the transformation for your specific router before it's passed in the `location` prop

PR #104, opened 2020-11-29
fix: FileWidget customRequest support

davecoates:
- Properly support customRequest along with progress updates
- Fix bug when multiple={true} and limit={1} (limit wasn't enforced)
- Fix issue where preview sometimes was not generated when it should have been

### DIFF
--- a/js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
+++ b/js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
@@ -233,10 +233,7 @@ export function useFileList(
  * @menu-group Widgets
  * @forward-ref
  */
-function FileWidget(
-    props: UploadWidgetProps<File, HTMLInputElement>,
-    ref: React.RefObject<Upload>
-): React.ReactElement {
+function FileWidget(props: UploadWidgetProps<File, HTMLInputElement>, ref): React.ReactElement {
     let {
         input,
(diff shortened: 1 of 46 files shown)


js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
```diff
@@ -161,15 +176,33 @@ export function useFileList(
...
+                        // libraries attach this, eg. antd-img-crop). Log a warning if name isn't available but
+                        // set it to `uid` so things continue to work (preview may not work properly).
+                        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+                        // @ts-ignore
+                        let name = value.name;
+                        if (!name) {
+                            console.warn(
+                                'No name set for uploaded file. Consider adding a `name` property if passing a `Blob`.'
+                            );
+                            name = `${uid}`;
```
fanglong-alliance: why not valueUid? (as in, this will rely entirely on internal uid instead of using the one from value.uid)

fanglong-alliance (approved):
looks good otherwise

js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
davecoates: good pickup, not intentional

PR #105, opened 2020-12-01
Fix/async choices reset behaviour

davecoates:


js-packages/@prestojs/ui-antd/src/widgets/SelectAsyncChoiceWidget.tsx
```diff
@@ -306,6 +315,27 @@ function SelectAsyncChoiceWidget<
         listOptions,
         retrieveOptions,
     });
+    const listRef = useRef(list);
+    listRef.current = list;
```
fanglong-alliance, line 331: is this line necessary?

davecoates: yeah needs to be set every render. so the first time it will be initialised and then set to same value (i just did this so type on `useRef` doesn't have to include `null`)

fanglong-alliance (approved):

PR #106, opened 2020-12-08
fix: SelectAsyncChoiceWidget loading & label glitches

davecoates:
- Fix case where selected values would flash off and on as search occured
- Fix issue where labels would disappear and id was shown instead as list
  is filtered

### DIFF
--- a/js-packages/@prestojs/ui-antd/src/widgets/SelectAsyncChoiceWidget.tsx
+++ b/js-packages/@prestojs/ui-antd/src/widgets/SelectAsyncChoiceWidget.tsx
@@ -150,7 +150,7 @@ export type SelectAsyncChoiceProps<T> = SelectProps<ValueType<T>> &
         retrieveOptions?: Record<string, any>;
     };
 
-type LabeledValue<T> = { key: T; label: React.ReactNode; found: boolean };
+type LabeledValue<T> = { key: T; label: React.ReactNode; found: boolean; missingLabel: boolean };
 
 /**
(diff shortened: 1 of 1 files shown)


js-packages/@prestojs/ui-antd/src/widgets/SelectAsyncChoiceWidget.tsx
```diff
@@ -375,10 +393,14 @@ function SelectAsyncChoiceWidget<
...
                 return;
             }
+            setLastValue([value]);
             return onChange(value.value);
         },
         [onChange]
     );
+
+    const missingLabel =
+        value && (Array.isArray(value) ? value.find(v => v.missingLabel) : value.missingLabel);
```
fanglong-alliance: `value.find(v => v.missingLabel)` would give you the `v`, not the label itself. do you mean 
`value.find(v => v.missingLabel).missinglabel`?

fanglong-alliance: ah, looking at how its used, is this only intended as a bool check? if thats the case might be better to use `.some` instead of `.find` (less confusing)

davecoates: we don't actually need the value, just need to know whether it exists. i've changed it to `isMissingLabel` and converted to boolean value

fanglong-alliance (approved):

PR #108, opened 2021-02-12
fix: FileWidget fixes and enhancements

davecoates:
- useFileList: refactor second argument to be options object. Add getThumbUrl
  to make customising thumbnail generation easier.
- fix bugs around multi uploads
- Make updateFileStatus easier to use by accepting UID and percent, status args.
  Previously to make changes you had to mutate an object and pass it through.
- Remove custom key property; we can use uid to achieve the desired outcome


js-packages/@prestojs/ui-antd/src/widgets/FileWidget.tsx
```diff
@@ -142,127 +163,151 @@ export function useFileList(
...
+                                name,
+                                size: value.size,
+                                type: value.type,
+                                originFileObj: value,
+                            };
+                            if (previewImage) {
+                                if (previewImage) {
+                                    promises.push(generatePreview(value, uploadFile));
                                 }
                             }
```
fanglong-alliance, line 250: lets make sure, really sure, that previewImage is truly? 😛

fanglong-alliance (approved):
lgtm w/ one tiny gotcha

PR #109, opened 2021-02-24
feature: add deleteViewModel parameter to viewModelCachingMiddleware

jasongi-alliance:
Allows you to specify to modify the model to delete for a DELETE endpoint middleware

### DIFF
--- a/js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
+++ b/js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
@@ -51,6 +51,15 @@ type ViewModelCachingOptions<T> = {
     getDeleteId?: ((context: MiddlewareContext<T>) => PrimaryKey) & {
         validateEndpoint?: (endpoint: Endpoint) => void;
     };
+    /**
+     * By default the mapping passed to `viewModelCacheMiddleware` is assumed to be the model to delete. In some
+     * cases you may want to return additional data from an endpoint and need to define a more advanced mapping - in
+     * these cases you can specify `deleteViewModel` as the model to delete and the mapping will be used to cache
+     * the result.
+     */
+    deleteViewModel?:
+        | ViewModelConstructor<any>
+        | (() => ViewModelConstructor<any> | Promise<ViewModelConstructor<any>>);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
```diff
@@ -208,6 +218,9 @@ export default function viewModelCachingMiddleware<ReturnT = any>(
                 if (!result) {
                     return {} as ReturnT;
                 }
+                if (!deleteViewModel) {
```
davecoates, line 249: is this condition necessary? previously if there was a result it would go through `cacheAndTransform` - it's just that it would use the same model as the delete occurred for

jasongi-alliance: This was required when I was looking at chaining. Now it's not necessarily required, but passing it through to cacheAndTransform seems a bit strange. If there is no deleteViewModel, then it's expecting either nothing or the response to be the model it just deleted. If it's the model you just deleted then it is kind of unexpected to delete it from the cache but then add it back in? Is there a case where that happens? Soft delete maybe?

davecoates: I guess you could return like a list of records, eg. if the deleted record changed ordering of other records

but I agree it's confusing and the previous behaviour wasn't really explicitly documented anyway. I'm happy with this - if you want response handled you have to specify `deleteViewModel`.

Can you update the documentation for the `viewModelMapping` to describe this, maybe:

```
* @param viewModelMapping The mapping to use for caching as described above. When the request method is `DELETE` this
 * is assumed to be the model that should be removed from the cache unless you provide `options.deleteViewModel`. If
 * `deleteViewModel` is provided then this value is used to cache the response of the delete call.
 ```

can you also add an example that shows deleting and handling the response?

Comment (jasongi-alliance) [reply js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts:249]: Done

File: js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts [hunk 2]
```diff
@@ -137,10 +146,38 @@ defaultGetDeleteId.validateEndpoint = (endpoint: Endpoint): void => {
  *
  * Each record instance created is also automatically added to the cache.
  *
+ * Usually when deleting a model, the endpoint doesn't return anything. So by default if the endpoint method is DELETE
+ * you should pass the model that will be deleted on a successful response.
+ *
+ * ```js
+ * const middleware = [viewModelCachingMiddleware(User)];
+ * // Response is empty:
+ * const userDelete = new ViewModelEndpoint(new UrlPattern('/users/:id/'), {
```
davecoates: oops - existing comments use `ViewModelEndpoint` but it's just `Endpoint` now. Could you replace all usages with `Endpoint`?

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
```diff
@@ -137,10 +146,38 @@ defaultGetDeleteId.validateEndpoint = (endpoint: Endpoint): void => {
...
+ * });
+ * ```
+ *
  * NOTE: If using with [paginationMiddleware](doc:paginationMiddleware) then this must come
  * before `paginationMiddleware`.
  *
- * @param viewModelMapping The mapping to use for caching as described above
+ * @param viewModelMapping The mapping to use for caching as described above. When the request method is `DELETE` this
+ * is assumed to be the model that should be removed from the cache unless you provide `options.deleteViewModel`. If
+ * deleteViewModel` is provided then this value is used to cache the response of the delete call.
```
davecoates: missing opening backtick around `deleteViewModel`

davecoates (changes requested):

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
jasongi-alliance: fixed

js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
jasongi-alliance: done

PR #110, opened 2021-02-28
fix: viewModelCache cache.getAll triggering an error on models with compound keys

liz-johnson-alliance:
Test2 has a compound key of id1, id2. You must provide an object mapping these fields to their values.
When cache.getAll calls cache.get, it uses a string value for pk eg. id2⁞6id1⁞5

This was triggering the error where it checks that compound keys are an object.

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
+++ b/js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
@@ -1488,7 +1488,8 @@ export default class ViewModelCache<
         >[];
         // Tracks if records have changed from what is stored in `lastRecords`
         let isRecordsSame = true;
-        for (const pk of this.cache.keys()) {
+        for (const recordCache of this.cache.values()) {
+            const pk = recordCache.recordPk;
             const record = this.get(pk, fieldNames as FieldNames[]);
             if (record) {
                 records.push(record);
--- a/js-packages/@prestojs/viewmodel/src/__tests__/ViewModelCache.test.ts
+++ b/js-packages/@prestojs/viewmodel/src/__tests__/ViewModelCache.test.ts
@@ -151,6 +151,10 @@ test('should validate pk(s)', () => {
     expect(() => Test2.cache.get({ id1: null }, ['name'])).toThrowError(
         'Test2 has a compound key of id1, id2. Missing value(s) for field(s) id1, id2'
     );
+
+    expect(Test2.cache.getAll(['id1', 'id2', 'name'])[0]).toBe(record2);
+    expect(Test2.cache.getAll(['name'])[0]).toBe(record2);
+
 });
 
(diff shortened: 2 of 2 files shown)


js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -1228,14 +1228,23 @@ export default class ViewModelCache<
         const isCompound = Array.isArray(pkFieldName);
         // typescript failed to narrow pkFieldName to array using isCompound here
         // for some reason...
-        if (Array.isArray(pkFieldName) && typeof pk !== 'object') {
+        if (Array.isArray(pkFieldName) && typeof pk !== 'object' && typeof pk !== 'string') {
```
davecoates: right now this will work because the cache key for a PK is then used directly below _and_ the key includes the names of the primary keys... but this could change (the key doesn't actually need to include this - it's an implementation detail). I'd like to avoid supporting this format in the interface for `get` and leave it as is.

The alternative fix is pretty simple - in `getAll` just change the loop to iterate over the caches instead of keys and extract the PK value from there (which will be an object if it's a compound key):

```
for (const recordCache of this.cache.values()) {
            const pk = recordCache.recordPk;
```

the rest should just work as is

js-packages/@prestojs/viewmodel/src/__tests__/ViewModelCache.test.ts
```diff
@@ -151,6 +151,9 @@ test('should validate pk(s)', () => {
     expect(() => Test2.cache.get({ id1: null }, ['name'])).toThrowError(
         'Test2 has a compound key of id1, id2. Missing value(s) for field(s) id1, id2'
     );
+
+    expect(Test2.cache.getAll(['id1', 'id2', 'name'])[0].id1).toBe(record2.id1)
```
davecoates: it should return the exact same record so can just be

```
expect(Test2.cache.getAll(['id1', 'id2', 'name'])[0]).toBe(record2);
```

also can you add case for just passing the 'name':

```
expect(Test2.cache.getAll(['name'])[0]).toBe(record2);
```

(should work - pks should be included automatically)

davecoates (changes requested):
thanks for looking at this 👍

js-packages/@prestojs/viewmodel/src/__tests__/ViewModelCache.test.ts
liz-johnson-alliance: I tried .toBe(record2) initially, but it didn't work. I'll try it again though and post the error.

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
liz-johnson-alliance: OK fixed

davecoates (approved):

PR #111, opened 2021-03-03
Fix/middleware order checks

davecoates:
Easier to review commit by commit - 7e8b76a remove a try/catch block so big diff for little change


fanglong-alliance:
lgtm

PR #115, opened 2021-03-18
detectBadPaginationMiddleware

davecoates:


doc-site/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 4000",
```
fanglong-alliance, line 6: whats this? o.O

davecoates: run the doc site on a different port (defaults to 3000... same as webpack server)

js-packages/@prestojs/util/src/pagination/CursorPaginator.ts
```diff
@@ -199,4 +200,36 @@ export default class CursorPaginator extends Paginator<
...
+        if (Array.isArray(decodedBody) || !decodedBody) {
+            return false;
+        }
+
+        // Cursor pagination responses should contain a next and previous link that includes
+        // a query parameter with name 'cursor' containing the next/previous cursor value. Results
+        // should be an array under the `results` key.
+        const isCursorLink = (link: any): boolean =>
+            typeof link == 'string' && 'cursor' in qs.parse(qs.extract(link));
+        if (isCursorLink(decodedBody.next) || isCursorLink(decodedBody.previous)) {
```
fanglong-alliance: what happens if both next & previous are null? (ie. paginated but single page)

shouldnt the middleware still treat it as paginated?

js-packages/@prestojs/util/src/pagination/PageNumberPaginator.ts
```diff
@@ -50,6 +53,12 @@ export default class PageNumberPaginator extends Paginator<
         if (typeof page === 'string') {
             return Number(page);
         }
+        if (page == null) {
```
fanglong-alliance, line 56: should this be a === just because?

also, yay

davecoates: want to catch either undefined or null

js-packages/@prestojs/util/src/pagination/CursorPaginator.ts
davecoates: yes good catch - existing code was buggy in this way

i've pushed a fix now

fanglong-alliance (approved):
lgtm

PR #117, opened 2021-04-06
fix: return initial state while query state is being updated

jasongi-alliance:
There is an issue where initial state isn't returned immediately after using the hook, this can be bad when you're using these params to make API calls. This is a fairly simple fix but doesn't account for the usecase where the params are wiped for whatever reason after the initial render.


davecoates:
Superseded by #118

PR #120, opened 2021-04-07
fix: range fields

fanglong-alliance:
draftish


js-packages/@prestojs/viewmodel/src/fields/RangeField.ts
```diff
@@ -19,15 +14,11 @@ interface Boundary<T> extends FieldProps<T> {
...
-    public lowerBound: T | null | undefined;
-    public upperBound: T | null | undefined;
     public separator: string;
 
-    constructor(values: Boundary<T>) {
-        super(values);
-        const { lowerBound, upperBound, separator = '~' } = values;
-        this.lowerBound = lowerBound;
-        this.upperBound = upperBound;
+    constructor(values: RangeFieldProps<T>) {
```
davecoates: should default to `= {}` (nothing is required - should be able to construct with eg. `new IntegerRangeField()`)

js-packages/@prestojs/ui-antd/src/widgets/RangeWidget.tsx
```diff
@@ -16,14 +16,33 @@ function RangeWidget<FieldValue, T extends HTMLElement, P>(
...
-                <Input placeholder={separator} disabled />
-                <InputWidget ref={upperRef} {...upperInput} {...rest} />
+                <InputWidget
+                    ref={lowerRef}
+                    defaultValue={value.lower}
+                    onChange={(v: any): void => onChange({ ...value, lower: v })}
+                    {...lowerInput}
+                    {...rest}
+                />
+                {separator}
```
davecoates: can you change this back to disabled `Input`?

```tsx
<Input className="presto-range-widget-separator" placeholder={separator} disabled />
```

js-packages/@prestojs/ui-antd/src/widgets/RangeWidget.tsx
```diff
@@ -16,14 +16,33 @@ function RangeWidget<FieldValue, T extends HTMLElement, P>(
...
+        upperInput,
+        separator = '-',
+        inputWidget: InputWidget,
+        input: { value = {} as Record<string, FieldValue>, onChange },
+        ...rest
+    } = props;
+    const { lowerRef, upperRef } = ref || {};
     return (
         <>
             <Input.Group compact>
```
davecoates: add className `className="presto-range-widget-wrapper"`

actually should also support `className` prop at top level I guess, eg. `className={cx("presto-range-widget-wrapper", className)}`

js-packages/@prestojs/ui-antd/src/widgets/RangeWidget.tsx
```diff
@@ -16,14 +16,33 @@ function RangeWidget<FieldValue, T extends HTMLElement, P>(
...
-                <InputWidget ref={lowerRef} {...lowerInput} {...rest} />
-                <Input placeholder={separator} disabled />
-                <InputWidget ref={upperRef} {...upperInput} {...rest} />
+                <InputWidget
+                    ref={lowerRef}
+                    defaultValue={value.lower}
+                    onChange={(v: any): void => onChange({ ...value, lower: v })}
+                    {...lowerInput}
+                    {...rest}
```
davecoates: it seems wrong that `rest` is passed through here and below. Seems like this should be used at the top level wrapper (if at all).

js-packages/@prestojs/ui-antd/src/widgets/RangeWidget.tsx
```diff
@@ -16,14 +16,33 @@ function RangeWidget<FieldValue, T extends HTMLElement, P>(
...
+    const { lowerRef, upperRef } = ref || {};
     return (
         <>
             <Input.Group compact>
-                <InputWidget ref={lowerRef} {...lowerInput} {...rest} />
-                <Input placeholder={separator} disabled />
-                <InputWidget ref={upperRef} {...upperInput} {...rest} />
+                <InputWidget
+                    ref={lowerRef}
+                    defaultValue={value.lower}
```
davecoates: why do we need to set `defaultValue`?

js-packages/@prestojs/ui-antd/src/widgets/RangeWidget.tsx
davecoates: can you add a class `presto-range-widget-lower` here? It should merge with `lowerInput.className` if it's set

js-packages/@prestojs/ui-antd/src/widgets/RangeWidget.tsx
```diff
@@ -16,14 +16,33 @@ function RangeWidget<FieldValue, T extends HTMLElement, P>(
...
+                    {...lowerInput}
+                    {...rest}
+                />
+                {separator}
+                <InputWidget
+                    ref={upperRef}
+                    defaultValue={value.upper}
+                    onChange={(v: any): void => onChange({ ...value, upper: v })}
+                    {...upperInput}
+                    {...rest}
```
davecoates: as above re: `rest`

can you add a class `presto-range-widget-uper` here? It should merge with `upperInput.className` if it's set

js-packages/@prestojs/ui/src/FieldWidgetInterface.ts
```diff
@@ -58,6 +58,8 @@ export interface WidgetProps<FieldValue, T extends HTMLElement> {
 export interface RangedWidgetProps<FieldValue, T extends HTMLElement, P> {
```
davecoates, line 61: can you add header comment so docs expand properties

```
/**
 * @expand-properties
 * /
```

js-packages/@prestojs/ui/src/FieldWidgetInterface.ts
```diff
@@ -58,6 +58,8 @@ export interface WidgetProps<FieldValue, T extends HTMLElement> {
...
     lowerInput: P;
     upperInput: P;
+    // the "input" from default form
+    input: InputProps<FieldValue, T> & { value?: Record<string, FieldValue> };
```
davecoates, line 73: can you document each key here please

should be format

```
/**
 * lowerInput docs
 */
lowerInput: P
/**
  * upperInput docs
  */
upperInput: P
...
```

you can run `cd doc-site && node extract-docs.js` and visit `localhost:4000` to see what docs look like when rendered

davecoates (changes requested):
~For styles can you create a `ui-antd/styles.less`:~

Can you create `RangeWidget.less` in same directory as `RangeWidget.tsx` with the below content. Then `ui-antd/styles.less` can import that. This way we can easily support only importing styles you need eg. using babel-plugin-import like we do for antd.

```less
@import (reference) '~antd/lib/style/index.less';

.presto-range-widget-wrapper.@{ant-prefix}-input-group {
  .presto-range-widget-upper {
    border-left-width: 0;

    &:hover, &:focus {
      border-left-width: 1px;
    }
  }
  .presto-range-widget-separator {
    width: 30px;
    border-left: 0;
    border-right: 0;
    pointer-events: none;
    margin-left: 1px;
  }

  .presto-range-widget-lower {
    border-right-width: 0;

    &:hover, &:focus {
      border-right-width: 1px;
    }
  }

  .presto-range-widget-separator {
    background-color: @input-bg;
  }
}
```

not entirely sure way we want to handle styles (have avoided it up until now) but for now this will do

then can import in project less file: `@import '~@prestojs/ui-antd/styles.less';`

(we won't build the less => css yet - if you want the styles have to use less)

js-packages/@prestojs/ui-antd/src/widgets/RangeWidget.tsx
davecoates: as discussed should set `value` instead

js-packages/@prestojs/ui-antd/src/widgets/RangeWidget.tsx
fanglong-alliance: yeah dropping that since we're asking people to customize via lowerInput/upperInput

js-packages/@prestojs/ui-antd/src/widgets/DateRangeWidget.tsx
```diff
@@ -20,8 +20,8 @@ function DateRangeWidget(
     ref: React.RefObject<DatePickerProps>
 ): React.ReactElement {
     const { lowerInput, upperInput, separator, ...rest } = props;
-    const { format: formatLower = 'MMMM Do YYYY', ...restLower } = lowerInput || {};
-    const { format: formatUpper = 'MMMM Do YYYY', ...restUpper } = upperInput || {};
+    const { format: formatLower = 'YYYY-MM-DD', ...restLower } = lowerInput || {};
+    const { format: formatUpper = 'YYYY-MM-DD', ...restUpper } = upperInput || {};
```
davecoates: i don't think we want to do this - this is shown to the user and YYYY-MM-DD is much less friendly/familiar than what was there

js-packages/@prestojs/ui-antd/src/widgets/DateWidget.tsx
```diff
@@ -22,7 +22,7 @@ function DateWidget(
     ref: React.RefObject<React.ClassicComponent<DatePickerProps, any>>
 ): React.ReactElement {
     const { input, ...rest } = props;
-    const { format = 'MMMM Do YYYY', ...restInput } = input;
+    const { format = 'YYYY-MM-DD', ...restInput } = input;
```
davecoates: as above

js-packages/@prestojs/ui-antd/src/styles.less
```diff
@@ -1 +0,0 @@
-@import './widgets/RangeWidget.less';
```
davecoates: we want to keep this file (just update import)

so in the end you can either import all styles (this file) or individual component styles

davecoates (changes requested):

PR #123, opened 2021-04-09
feature: AntdUiProvider

davecoates:


fanglong-alliance:
we're missing Calendar; think we should have a default support for Calendar too? (yes not using it currently, otoh might want to do it now so future uses would be safer)

js-packages/@prestojs/ui-antd/src/AntdUiProvider.tsx
```diff
@@ -0,0 +1,200 @@
...
+        if (!this.components.DatePicker) {
+            throw new Error(`To use components that use DatePicker you must first configure @prestojs/ui-antd so it knows the type to use. To set the default moment based datepicker add the following to your app before any widgets are used:
+
+    import { DatePicker } from 'antd';
+
+    <AntdUiProvider datePickerComponent={DatePicker}>...</AntdUiProvider>
+
+See https://prestojs.com/docs/ui-antd/ for instructions on setting up @prestojs/ui-antd and 
+https://ant.design/docs/react/replace-moment#DatePicker for options in customising the DatePicker
+`);
```
fanglong-alliance, line 64: i think we should mention that moment is now legacy and generally discouraged
(ant's not doing that for major version backwards compatibility reason but we dont have that burden)

fanglong-alliance:
just a note, would need to change rangefields once they're in as well. 
lgtm

davecoates:
> just a note, would need to change rangefields once they're in as well.
> lgtm

ah yep - forgot to mention that (didn't want to introduce conflicts so intentionally left it out)

davecoates:
> we're missing Calendar; think we should have a default support for Calendar too? (yes not using it currently, otoh might want to do it now so future uses would be safer)

I'm not sure there's any benefit doing it now - may even mislead into thinking it's available somewhere. Would also mean project pulls in Calendar even though never used anywhere (we'd use lazy load but still generates another bundle for no reason).

PR #126, opened 2021-04-16
feature: decimal range fields and listfield widget mapping

fanglong-alliance:


js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
```diff
@@ -99,10 +101,19 @@ export default function getWidgetForField<
...
     // Couldn't work out what to type this as so field.constructor was accepted
-    const widget: FieldWidgetType<any, any> | null | undefined =
-        field.choices || field.asyncChoices
-            ? choicesMapping.get(fieldClassName) || mapping.get(fieldClassName)
-            : mapping.get(fieldClassName);
+    const getWidget = (clsName: string): FieldWidgetType<any, any> | null | undefined => {
+        if (clsName === 'ListField') {
+            if (field.choices) {
+                return React.lazy(() => import('./widgets/SelectChoiceWidget'));
+            } else {
```
davecoates: this doesn't seem safe - why the assumption that this is `asyncChoices`? This looks like it will break if it's not, eg. `new ListField({ childField: new IntegerField() })`

fanglong-alliance: what should be the default for ListField you think? we were not catering for it before - shall we just raise if both choices & asyncChoices are missing?

js-packages/@prestojs/ui-antd/src/widgets/PasswordWidget.tsx
```diff
@@ -0,0 +1,23 @@
+import { WidgetProps } from '@prestojs/ui';
+import { Input } from 'antd';
+import React from 'react';
+
+type PasswordWidgetProps = WidgetProps<string, HTMLInputElement>;
+
+/**
+ *
+ * See [Input](https://ant.design/components/input/) for props available
```
davecoates: just to make it clear that it renders `Input.Password` can you link to that?

```
[Input.Password](https://ant.design/components/input/#Input.Password)
```

davecoates (changes requested):

PR #129, opened 2021-04-22
AntdUiProvider - remove default props

davecoates:


vbudovski-alliance:
LGTM

doc-site/components/api-docs/ReturnType.js
```diff
@@ -39,25 +39,36 @@ export default function ReturnType({ signature }) {
...
-                        {!isConstructor && c && (
-                            <ParameterTable parameters={c} nameHeader="Key" isReturnType />
-                        )}
-                        {i !== returnTypes.length - 1 && <p className="my-10 text-center">OR</p>}
-                    </React.Fragment>
-                );
-            })}
+            {signature.comment?.tagsByName?.['return-type-name'] ? (
+                <p className="text-blue-400">
+                    {signature.comment?.tagsByName?.['return-type-name']}
```
fanglong-alliance: nitpick: if you're here you should not need all ?s 🤪

django-testbed/frontend/src/App.js
```diff
@@ -1,16 +1,20 @@
...
 // eslint-disable-next-line import/extensions
 import './styles/global.less?no-css-modules';
-import TimePicker from './TimePicker';
+
 import UserListView from './views/UserListView';
 
+const DatePicker = React.lazy(() => import('./DatePicker'));
+const TimePicker = React.lazy(() => import('./TimePicker'));
+const FormItemWrapper = React.lazy(() => import('@prestojs/ui-antd/build/FormItemWrapper'));
+const FormWrapper = React.lazy(() => import('@prestojs/ui-antd/build/FormWrapper'));
```
fanglong-alliance: hmm, why importing from build?

js-packages/@prestojs/ui-antd/src/AntdUiProvider.tsx
```diff
@@ -137,52 +135,127 @@ export function useAntdUiConfig(): AntdUiConfig {
...
- *     <AntdUiProvider>
+ *     <AntdUiProvider datePickerComponent={datePickerComponent} timePickerComponent={timePickerComponent} >
  *        <YourApp />
  *     </AntdUiProvider>
  *   );
  * }
  * ```
  *
- * All props from [UiProvider](doc:UiProvider) are supported and default to:
+ * See the antd [replace moment](https://ant.design/docs/react/replace-moment#DatePicker) docs for defining components
```
fanglong-alliance: this # looks pretty suspicious

js-packages/@prestojs/ui-antd/src/AntdUiProvider.tsx
```diff
@@ -137,52 +135,127 @@ export function useAntdUiConfig(): AntdUiConfig {
...
+} & UiProviderProps;
+
+/**
+ * Version of [UiProvider](doc:UiProvider) that supports some extra options specific to antd.
+ *
+ * If you use date or time widgets in the system you should pass through the `datePickerComponent` and `timePickerComponent`
+ * props. Without this when [DateWidget](doc:DateWidget), [TimeWidget](doc:TimeWidget) or other date/time based components
+ * are used they will throw an error. The reason for this is to configure how you want dates to be handled. The default
+ * provided by antd is to use momentjs but you may wish to switch this to date-fns or dayjs. To use the default (moment)
+ * based components do:
```
fanglong-alliance: (same comment as b4: we might want to use stronger word to recommend against moment)

doc-site/components/api-docs/ReturnType.js
davecoates: nope!

django-testbed/frontend/src/App.js
davecoates: because this is run locally not from a proper build. when we do a proper build & publish it copies stuff such that the published directory serves from the root.

Here the best we can do is import from `build`... but even that will only work if you have run a local build. I'll probably just change this to remove the `React.lazy` - i was just testing stuff and it's just the testbed so not critical

PR #130, opened 2021-04-26
fix/field widget selection fixes

davecoates:


js-packages/@prestojs/ui-antd/src/widgets/ChoicesWidget.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+    if (widgetType !== 'select' && asyncChoices) {
+        throw new Error(
+            `When 'asyncChoices' is provided 'widgetType' must be 'select', received: ${widgetType}`
+        );
+    }
+    if (asyncChoices) {
+        if (multiple != null && asyncChoices.multiple !== multiple) {
+            throw new Error(
+                `When 'asyncChoices' is specified 'multiple' does not need be specified. Got mismatch - 'asyncChoices.multiple' = ${asyncChoices.multiple.toString()} but 'multiple' = ${multiple.toString()}`
```
fanglong-alliance, line 96: feels like this should be a warning outside (so people would learn to stop doing this w/o having to make the mistake 1st)

fanglong-alliance:
kgtm

PR #131, opened 2021-04-27
fix: fix TimeFormatter on safari

jasongi-alliance:
Safari doesn't support Date.parse(`0999-04-11 11:11`), it will return NaN

It will however support Date.parse(`0999-04-11T11:11`), so this will fix that.


davecoates:
thanks!

PR #139, opened 2021-06-17
feature: paginationMiddleware 'path' option

davecoates:


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Add [CheckboxChoicesWidget](https://prestojs.com/docs/ui-antd/widgets/CheckboxChoicesWidget/)
 * Add [ChoicesWidget](https://prestojs.com/docs/ui-antd/widgets/ChoicesWidget/)
 * Add [parseTime](https://prestojs.com/docs/util/parseTime/) and [formatTime](https://prestojs.com/docs/util/formatTime/)
+* Add [path](https://prestojs.com/docs/rest/paginationMiddleware/) option to `paginationMiddleware`. This allows pagination to be handled on a sub-key of the data.
```
vbudovski-alliance: I don't love the name... Maybe resultKey ?

js-packages/@prestojs/rest/src/paginationMiddleware.ts
```diff
@@ -4,6 +4,10 @@ import {
     PaginatorInterface,
     PaginatorInterfaceClass,
 } from '@prestojs/util';
+import cloneDeep from 'lodash/cloneDeep';
+import get from 'lodash/get';
+
```
vbudovski-alliance, line 9: I'm surprised the sorting plugin didn't fix the space. Is it configured for Presto?

CHANGELOG.md
davecoates: key sounds like it's only a direct key and not a dotted path... `resultPath`?

vbudovski-alliance: Hmm... maybe a very explicit `dottedResultKey`/`dottedResultPath`/`dottedResultProperty`?

vbudovski-alliance: Also, result or response?

davecoates: it can be dotted or not though (and most cases will be not)

`result` because that's what we call it after the body has been decoded from response and transformed

vbudovski-alliance: Go with `resultPath` then.

PR #140, opened 2021-06-21
Endpoint baseurl + middleware enhancements + batchMiddleware

davecoates:


js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -117,7 +131,50 @@ export type ExecuteReturnVal<T> = {
...
-export type MiddlewareContext<T> = {
+/**
+ * This can be passed to `next` in middleware to skip all subsequent middleware in the chain, the call
+ * to `fetch` and resolve instead to the `promise` passed in.
+ *
+ * This is useful for middleware like [dedupeInFlightRequestsMiddleware](doc:dedupeInFlightRequestsMiddleware) where
+ * if a duplicate request is detected it just needs to wait for the original request to finish and return the
+ * same Response object.
+ *
+ * It can also be used to change how `fetch` is performed like in `batchMiddleware`.
```
levic:
```suggestion
 * It can also be used to change how `fetch` is performed (`batchMiddleware` uses this).
```

js-packages/@prestojs/rest/src/Endpoint.ts
```diff
@@ -117,7 +131,50 @@ export type ExecuteReturnVal<T> = {
...
+ *
+ * For example this just calls a completely different URL and uses the response object from that:
```
levic, line 144: Possibly worth mentioning that will completely bypasses any subsequent middleware (including response handling). If there is any security-related middleware it should come before this is called.

I'm assuming that the middleware that comes before this would still get a chance to manipulate the response returned from this?

davecoates: yep everything before it can modify the response

vbudovski-alliance (approved):

PR #146, opened 2021-07-21
Fix/pagination warnings and inference

davecoates:


js-packages/@prestojs/rest/src/viewModelCachingMiddleware.ts
```diff
@@ -23,6 +30,18 @@ function cacheDataForModel<T extends ViewModelConstructor<any>>(model: T, data):
         model.cache.addList(records);
         return records;
     }
+    if (!isPkSet(data, model.pkFieldName)) {
+        console.warn(
+            `Data received by viewModelCachingMiddleware does not look a single record or list of records. If the data is nested under a key make sure you pass the mapping to 'viewModelCachingMiddleware'.
+
+     * If the response is for a list of records check that it is an array. If the data is paginated make sure the endpoint includes 'paginationMiddleware'.
```
levic, line 37: I don't think it's completely clear what's supposed to be happening if you are doing this and forget the pagination middleware.

Instead of 

`If the data is paginated make sure`

maybe this? (not completely happy with the wording but is the idea)

`If the array is contained within a pagination envelope make sure`

levic (approved):
Looks good to me, the console message would have saved quite a bit of time doing the training project

PR #151, opened 2021-11-12
feature: field boundRecord value

davecoates:


jasongi-alliance (approved):
Looks fine but also would be good to fix those lint errors before merging

PR #154, opened 2022-01-20
Rewrite ViewModelCache

davecoates:
This implementation is greatly simplified (and about half as long) but the ViewModelCache file has been essentially rewritten - view the whole file rather than trying to read the diff. This resolves some long standing bugs (see test cases that were previously skipped) and makes it easier to work with for any extensions. 

ViewModelFactory has also had large parts rewritten. Notably there is now a base class that is always extended rather than being entirely dynamic. This file might be easier to be viewed in its entirety rather than a diff but a lot of it is old code just moved around.

Other notable changes
- ViewModelFactory no longer supports implicit primary keys. You always must define this yourself.
- You can now specify the cache to use on a relation field (defaults to the ViewModel.cache). This is mainly useful for testing.
- Types now verify field names across relations.
    - These _don’t_ work across circular relations but we can live with that I think
- Types are generally much more comprehensive for caching and factory
- Add [tsd](https://github.com/SamVerschueren/tsd) for typescript testing


js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -40,643 +45,195 @@ function withEnableListeners<T>(run: () => T): T {
...
+    cacheListeners: Map<
+        ViewModelFieldPaths<ViewModelClassType>,
+        ChangeListener<PartialViewModel<ViewModelClassType>>[]
+    >;
     /**
-     * Take a cache key generated with `getCacheKey` and return the list of fields
-     *
-     * Must be kept in sync with `getCacheKey`/`getFieldNameCacheKey`
+     * When a record has relation fields it needs to listen for changes on those records so it can
+     * update itself and notify it's direct listeners. This Map keeps track of those listeners.
```
fanglong-alliance: possible to rewrite this comment? had to read it quite a few times to get what its saying (with that tingling uncertainty)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -40,643 +45,195 @@ function withEnableListeners<T>(run: () => T): T {
...
+    viewModel: ViewModelClassType;
+    onAnyChange: () => void;
+    /**
+     * Contains all keys in `cache` in descending order of last insert/update
      *
-     * @private
+     * This is used to choose the most recent record when a new record with a subset
+     * of the fields needs to be created. For example if there are 2 existing records
+     * created in this order:
+     * { id: 1, firstName: 'Jon', lastName: 'Doe', active: true }
```
fanglong-alliance: would suggest change the `active` key in this example to sth else as active could have other meanings (ie. its easy for this to be seen as not part of record itself)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -40,643 +45,195 @@ function withEnableListeners<T>(run: () => T): T {
...
+        this.cacheListeners = new Map();
+        this.relationListenerUnsubscribe = new Map();
+        this.viewModel = viewModel;
+        this.cache = new Map();
+        this.recordPk = pk;
+        this.onAnyChange = onAnyChange;
     }
 
     /**
      * Set a value for the specified key notifying any listeners
```
fanglong-alliance: * Set a value for the specified key_, then_ notify any listeners
?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -40,643 +45,195 @@ function withEnableListeners<T>(run: () => T): T {
...
-                                    this.setValueForKey(fieldsKey, newRecord);
-                                }
-                            }
-                        },
-
-                        false
-                    )
-                );
-            }
+        let before = this.cache.get(key) || null;
```
fanglong-alliance, line 133: so even if the value is the same, the lastUpdatedKeys order would still change - this could get a mention in doc if its intended

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -40,643 +45,195 @@ function withEnableListeners<T>(run: () => T): T {
...
-        const { nonRelationFieldNames: otherNonRelationFieldKeys } = this.getRelationFields(
-            fieldNames
-        );
-        const { relations, nonRelationFieldNames } = this.getRelationFields(cacheFieldNames);
-        if (isSubset(flattenFieldPath(cacheFieldNames), fieldNames)) {
-            if (record instanceof RecordPointer) {
-                record = record.record;
-            }
-            return new RecordPointer(this.cache.get(key), record);
+        if (before && value && isShallowEqual(before, value)) {
```
fanglong-alliance: uh, would isShallowEqual return true if one of them is null but the other is not? 
as both null cases are covered above already

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -40,643 +45,195 @@ function withEnableListeners<T>(run: () => T): T {
...
-                    continue;
+    get(key: ViewModelFieldPaths<ViewModelClassType>): PartialViewModel<ViewModelClassType> | null {
+        const record = this.cache.get(key);
+        if (record) {
+            return record;
+        }
+        // record doesn't already exist for key - see if it's a subkey of existing entries
+        for (const cacheKey of this.lastUpdatedKeys) {
+            const record = this.cache.get(cacheKey);
+            if (record && cacheKey.isSubset(key, true)) {
```
fanglong-alliance, line 201: performance: is the .get heavier or isSubset heavier? asking coz order of execution here's a bit unintuitive

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -40,643 +45,195 @@ function withEnableListeners<T>(run: () => T): T {
...
+                const data = key.nonRelationFieldNames.reduce((acc, path) => {
+                    acc[path] = record[path];
+                    return acc;
+                }, {});
+                for (const [relationFieldName, relationFieldPaths] of Object.entries(
+                    key.relations
+                )) {
+                    const relationField = this.viewModel.getField(
+                        relationFieldName
+                    ) as BaseRelatedViewModelField<any, any, any>;
```
fanglong-alliance, line 213: (ugh, whats the line width setting we're using in presto..)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -40,643 +45,195 @@ function withEnableListeners<T>(run: () => T): T {
...
+                    const idOrIds = record[relationField.sourceFieldName];
+                    // If the id is null (or empty array for many-to-many) then there is nothing to resolve - set to null/[]
+                    if (idOrIds == null || (Array.isArray(idOrIds) && idOrIds.length === 0)) {
+                        data[relationFieldName] = relationField.many ? [] : null;
+                    } else {
+                        if (Array.isArray(idOrIds)) {
```
fanglong-alliance: it feels like the better check to use is `relationField.many`. are there cases that they'll  otherwise still be an array?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -686,94 +243,153 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
+        key: ViewModelFieldPaths<ViewModelClassType>,
+        baseRecord: PartialViewModel<ViewModelClassType>
+    ): PartialViewModel<ViewModelClassType> | null {
+        const newData = { ...baseRecord._data } as Record<string, any>;
+        for (const [relationFieldName, relationFieldPath] of Object.entries(key.relations)) {
+            const relationField = this.viewModel.getField(
+                relationFieldName
+            ) as BaseRelatedViewModelField<any, any, any>;
+            const idOrIds = baseRecord[relationField.sourceFieldName];
+            if (idOrIds != null) {
```
fanglong-alliance: minor: this's similar to one above so it might make sense to line up the logic? (ie. check idOrIds being empty array at this level & set newData[] as so to avoid hitting getList down below)

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -686,94 +243,153 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
         return true;
     }
 
     /**
-     * Add a listener for any changes, additions or deletions for the specified field names
-     * @param fieldNames field names to listen to any changes for. See [Field notation](#Field_notation) for supported format.
-     * @param listener Function to call with any changes
+     * Given a `baseRecord` with no related fields resolved (ie. we just have the id's, not the nested records) construct
+     * a record with all the nested records attached if those records exist in the cache.
+     *
```
fanglong-alliance, line 292: can use a brief mentioning here like '...if all of those records exist in the cache, otherwise return null'

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -686,94 +243,153 @@ class RecordCache<ViewModelClassType extends ViewModelConstructor<any, any, any>
...
+        let registeredPaths = this.relationListenerPaths.get(record);
+        if (!registeredPaths) {
+            registeredPaths = [];
+            this.relationListenerPaths.set(record, registeredPaths);
+        }
+        if (registeredPaths.includes(key)) {
+            return;
+        }
+        registeredPaths.push(key);
+        this.relationListenerUnsubscribe.get(key)?.forEach(unsub => unsub());
```
fanglong-alliance, line 360: why would there? think i might be missing sth - can i get an example?

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -1007,25 +668,59 @@ const defaultListenerBatcher = {
...
+    private isInstanceOfModel(a: any): a is PartialViewModel<ViewModelClassType> {
+        return a instanceof this.viewModel;
+    }
+
+    /**
+     * Get the cache key to use into for the primary key. Handles compound keys.
+     */
+    private getPkCacheKey(pk: ExtractPkFieldParseableValueType<ViewModelClassType>): string {
+        if (typeof pk === 'object') {
+            return this.viewModel.pkFieldNames.map(fieldName => pk[fieldName]).join('⁞');
```
fanglong-alliance: 😆 👍

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
```diff
@@ -1527,6 +991,78 @@ export default class ViewModelCache<
...
+     * @param records List of existing ViewModel records to get latest cache version for
+     * @param removeNulls whether to remove entries that have no record in the cache. Defaults to true.
+     * @returns an array of the cached records. Any records not found will be in the array as a null value if `removeNulls` is false otherwise they will be removed.
+     **/
+    getList<T extends FieldPath<ViewModelClassType>, RemoveNullsT extends boolean = true>(
+        records: PartialViewModel<ViewModelClassType, T>[],
+        removeNulls?: RemoveNullsT
+    ): RemoveNullsT extends true
+        ? PartialViewModel<ViewModelClassType, T>[]
+        : (PartialViewModel<ViewModelClassType, T> | null)[];
```
fanglong-alliance, line 1072: my eyes! 🤣
 
perhaps rearrange these 3 to make it a bit clearer? (mainly the 2nd overload)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -1,82 +1,306 @@
...
+
+type ExtractPkFields<
+    FieldMappingType extends FieldsMapping,
+    PkFieldType extends ExtractPkFieldTypes<FieldMappingType>
+> = PkFieldType extends string
+    ? Record<PkFieldType, FieldMappingType[PkFieldType]>
+    : {
+          [K in keyof FieldMappingType as K extends PkFieldType[number]
+              ? K
+              : never]: FieldMappingType[K];
```
fanglong-alliance, line 226: (this's really not pretty🤣 : wth's `:never]:`? )

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -1,82 +1,306 @@
...
+            ', '
+        )}.\n Missing fields: ${missingFieldNames.join(', ')}`;
+        if (missingRelations.length > 0) {
+            err +=
+                '\nThe following relations had no data so the associated fields could not be retrieved:\n';
+            err += missingRelations.map(
+                ([relationName, relationFieldNames]) =>
+                    ` The relation '${relationName}' for fields '${relationFieldNames.join(', ')}'`
+            );
+        }
```
fanglong-alliance, line 271: would the msg look funny if the only missing parts are missingRelations?

fanglong-alliance: ah, nvm - looks like wont happen w/ usage below

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -85,45 +309,322 @@ export type ViewModelInterface<
...
+                this as ViewModelInterface<any, any>,
+                assignedFields,
+                fieldNames as FieldPath<ViewModelConstructor<FieldMappingType, PkFieldType>>[],
+                missingFieldNames,
+                missingRelations
+            );
+        }
+
+        for (const fieldName of nonRelatedFieldNames) {
+            // TODO: Unclear to me if this needs to call a method on the Field on not. Revisit this.
```
fanglong-alliance: just checkimng that this todo would stay as todo for now?

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -85,45 +309,322 @@ export type ViewModelInterface<
...
+        for (const [key, value] of Object.entries(data)) {
+            const field = fields[key];
+            if (field) {
+                assignedData[key] = field.normalize(value);
+                if (field instanceof BaseRelatedViewModelField) {
+                    // Convert null to empty array for many fields
+                    if (field.many && !assignedData[key]) {
+                        assignedData[key] = [];
+                    }
+                    // TODO: Make doing this part of interface rather than special case?
```
fanglong-alliance: (todo)

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -85,45 +309,322 @@ export type ViewModelInterface<
...
+                        : assignedData[key]?._key;
+                    const hasValue = Array.isArray(pkOrPks)
+                        ? pkOrPks.length > 0 && data[field.sourceFieldName]?.length > 0
+                        : pkOrPks != null && data[field.sourceFieldName];
+                    if (hasValue && !isEqual(data[field.sourceFieldName], pkOrPks)) {
+                        const { name, sourceFieldName } = field;
+                        const pk = Array.isArray(pkOrPks) ? pkOrPks.join(', ') : pkOrPks;
+                        console.warn(
+                            `Related field ${name} was created from nested object that had a different id to the source field name ${sourceFieldName}: ${data[sourceFieldName]} !== ${pk}. ${pk} has been used for both.`
+                        );
```
fanglong-alliance, line 533: can we add the model.name to this warning? to make it a bit clearer

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -132,33 +633,90 @@ export type ViewModelInterface<
...
+        ? FieldMappingType[PkFieldType]['__fieldValueType']
+        : {
+              [K in keyof FieldMappingType as K extends PkFieldType[number]
+                  ? K
+                  : never]: FieldMappingType[K]['__fieldValueType'];
+          } {
+        const { pkFieldName } = this._model;
+        if (Array.isArray(pkFieldName)) {
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
```
fanglong-alliance: can use some text here explaining why (thought we're doing this everywhere else?)

js-packages/@prestojs/viewmodel/src/fieldUtils.ts
```diff
@@ -0,0 +1,244 @@
+import { BaseRelatedViewModelField } from './fields/RelatedViewModelField';
+import { FieldPath, FieldPaths, InvalidFieldError, ViewModelConstructor } from './ViewModelFactory';
+
+const expandedFields = new Map<ViewModelConstructor<any, any>, FieldPath<any>[]>();
+
+/**
+ * For a model expand the fields according to the rules for the '*' specification.
+ *
+ * This includes all field names apart from relation fields (but it does include the
+ * relation `sourceFieldName`).
```
fanglong-alliance: ^but

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
fanglong-alliance: looks like ⁞ was defined as a const in fieldUtils - should use here

js-packages/@prestojs/viewmodel/src/fieldUtils.ts
```diff
@@ -0,0 +1,244 @@
...
+                if (fieldPath === '*') {
+                    expandStarFields(modelClass).forEach(fieldName =>
+                        resolvedFieldNames.add(fieldName)
+                    );
+                } else if (Array.isArray(fieldPath)) {
+                    const [relatedFieldName, ...rest] = fieldPath;
+                    const field = modelClass.getField(relatedFieldName);
+                    if (!(field instanceof BaseRelatedViewModelField)) {
+                        throw new InvalidFieldError(
+                            `Field '${relatedFieldName}' does not extend BaseRelatedViewModelField.`
```
fanglong-alliance: also print the model name here?

js-packages/@prestojs/viewmodel/src/fieldUtils.ts
```diff
@@ -0,0 +1,244 @@
...
+                        resolvedFieldNames.add(fieldPath);
+                    }
+                }
+            }
+        }
+        const f = [...resolvedFieldNames].map(f => (Array.isArray(f) ? f.join('.') : f));
+        f.sort();
+        // See if it already exists
+        for (const viewModelFieldPaths of this.cache.values()) {
+            if (viewModelFieldPaths.key === f.join(CACHE_KEY_FIELD_SEPARATOR)) {
```
fanglong-alliance: consider to do a join outside? just to make it a tiny bit faster

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: ah yeah dunno why did this - should just be able to do the isShallowEqual check

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: `isSubset` i imagine - `get` is just a `Map.get` call

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: printWidth of 100 - same as all our projects :P

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: nup - will change

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: I've reversed the condition:

```
if (idOrIds == null || (Array.isArray(idOrIds) && idOrIds.length === 1)) {
    newData[relationFieldName] = relationField.many ? [] : null;
} else {
    ...
}
````

File: js-packages/@prestojs/viewmodel/src/ViewModelCache.ts (hunk 11 above)
Comment (davecoates) [reply js-packages/@prestojs/viewmodel/src/ViewModelCache.ts:360]: I'll add this comment

```
        // This may already be set for a previous call to this function for a different record (ie.
        // a previous version of the record with different related record id(s)). In that case remove
        // the old listeners and add new ones to ensure they match the latest related record id(s).
```

js-packages/@prestojs/viewmodel/src/ViewModelCache.ts
davecoates: hmm don't see how i can re-arrange it to make any meaningful difference

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: `never` will mean only fields that extend `PkFieldType[number]` will be included

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: i'll remove it - it's been there since original version

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
fanglong-alliance: yeah what i mean's prettier did a &^@~&!@ job there (ie a side comment hence its entirely inside a () pair) that it caused the line to begin with `:never]:` which gave me a pause 😆

fanglong-alliance (approved):

js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
davecoates: ahhh my bad i thought you meant what does it mean

PR #160, opened 2022-02-25
fix: resolve bug with nested related data

davecoates:
See changelog for example of what bug was


js-packages/@prestojs/viewmodel/src/ViewModelFactory.ts
```diff
@@ -635,6 +641,137 @@ export class BaseViewModel<
...
+                    return acc;
+                },
+                []
+            );
+            if (relatedRecords.length === 0) {
+                continue;
+            }
+            if (relation.many) {
+                console.log(relatedRecords);
+                // relatedRecords = relatedRecords[0];
```
callumanderson-alliance: debug?

davecoates: oops - ta

PR #166, opened 2022-05-29
React18 / StrictMode Support 

davecoates:
There were changes required to conform to requirements imposed by StrictMode


davecoates:
Resolves #165

package.json
```diff
@@ -78,10 +80,14 @@
...
     "@babel/preset-typescript": "^7.16.0",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "^18.1.0",
+    "react-dom": "^18.1.0",
     "wsrun": "^5.2.0"
   },
+  "resolutions": {
+    "@types/react": "^18.0.9",
+    "@types/react-dom": "^18.0.5"
```
vbudovski-alliance, line 89: Did you intend to do this?

davecoates: yep. without it seemed to get different versions which caused things to break

vbudovski-alliance: Better to pin to 18.1 maybe?

js-packages/@prestojs/doc/package.json
```diff
@@ -19,8 +19,8 @@
     "clean": "rm -rf ./build && rm -rf tsconfig.tsbuildinfo"
   },
   "peerDependencies": {
-    "react": "*",
-    "react-dom": "*",
+    "react": ">= 16.12.0 < 19",
```
vbudovski-alliance, line 22: Possibly just bump version and highlight breaking change React 18+?

vbudovski-alliance:
Couple of questions, but LGTM @davecoates

package.json
davecoates: Dunno how the versioning is done but it's not the same, see https://www.npmjs.com/package/@types/react

js-packages/@prestojs/doc/package.json
davecoates: I mainly wanted to avoid this so existing projects we have can upgrade presto for any bugfixes without also having to upgrade React18 at same time. This should work fine with the caveat that presto will no longer avoid the 'setState after unmount' warnings for you

PR #167, opened 2022-06-01
feature: add getWidgetProps, getFormatterProps to Field

davecoates:
This provides way for component specific props to make it's way from Field definition
to widget or formatter components.

This serves to primary purposes:
1) Allow a field like CharField to pass 'maxLength' to the component without getWidgetForField
   having to know the prop exists (previous solution).
2) Make it easy for end user to pass through arbitrary props when defining the model
   rather than having to pass it at each form/formatter usage.

### DIFF
--- a/js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
+++ b/js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
@@ -167,9 +167,9 @@ export default function getWidgetForField<
             const finalWidget = Array.isArray(_widget) ? _widget[0] : _widget;
             const props = Array.isArray(_widget) ? _widget[1] : {};
             const finalProps = {
+                ..._field.getWidgetProps(),
                 ...props,
                 ...extraProps,
-                ..._field.getWidgetProps(),
             };
             // Only set this when necessary to avoid passing props through
             // with undefined value that may make it's way through to the DOM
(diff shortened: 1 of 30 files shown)


js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
```diff
@@ -171,6 +169,7 @@ export default function getWidgetForField<
             const finalProps = {
                 ...props,
                 ...extraProps,
+                ..._field.getWidgetProps(),
```
fanglong-alliance: should it be the last here or first (or at least before extraProps)?

js-packages/@prestojs/ui/src/getFormatterForField.ts
```diff
@@ -99,11 +97,14 @@ export default function getFormatterForField<
...
         if (f.choices) {
             if (Array.isArray(w)) {
-                return [w[0], { ...w[1], choices: f.choices }];
+                return [w[0], { ...w[1], choices: f.choices, ...f.getFormatterProps() }];
             } else {
-                return [w, { choices: f.choices }];
+                return [w, { choices: f.choices, ...f.getFormatterProps() }];
             }
         } else {
+            if (w) {
```
fanglong-alliance: hmm, when would w be falsy? it also looks like its assumed if f is a choices

davecoates: hmm yeah looks like you are right - i'll add missing types to make it more obvious

js-packages/@prestojs/ui-antd/src/getWidgetForField.ts
davecoates: ah yeah good one - should come first so can be overriden

PR #169, opened 2022-06-10
fix: DateField/DateTimeField isEqual

### LINKED ISSUE #157: DateTimeField/DateField isEqual
Current these fields don't do equality checks on `Date` objects. We probably should:

```
    isEqual(a?: Date, b?: Date): boolean {
        if (!a || !b) {
            return false;
        }
        return a === b || a.getTime() === b.getTime();
    }
```

davecoates:
- DateField and DateTimeField now consider dates that represent the same
  date or datetime as equal
- This is important for caching otherwise a record with these fields will
  never be considered equal (eg. when record comes from API endpoint via
  viewModelCachingMiddleware it will construct an instance of the viewmodel
  then do an isEqual check to see if it differs from existing record)

Resolves #157

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/fields/DateField.ts
+++ b/js-packages/@prestojs/viewmodel/src/fields/DateField.ts
@@ -22,15 +22,17 @@ export default class DateField extends Field<Date, string | Date> {
     }
 
     isEqual(a?: Date, b?: Date): boolean {
+        if (a === b) {
+            return true;
+        }
         if (!a || !b) {
             return false;
         }
         return (
-            a === b ||
             // We only compare date components as there's technically no time component for DateField
-            (a.getFullYear() === b.getFullYear() &&
-                a.getMonth() === b.getMonth() &&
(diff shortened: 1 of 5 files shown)


js-packages/@prestojs/viewmodel/src/fields/DateField.ts
```diff
@@ -21,6 +21,19 @@ export default class DateField extends Field<Date, string | Date> {
         return new Date(value);
     }
 
+    isEqual(a?: Date, b?: Date): boolean {
+        if (!a || !b) {
+            return false;
```
fanglong-alliance, line 29: should it be considered equal if both values are none?

fanglong-alliance:
otherwise lgtm

js-packages/@prestojs/viewmodel/src/fields/DateField.ts
davecoates: 🤦‍♂️  thank you - definitely

PR #173, opened 2022-08-11
feature: async choices parseValue

davecoates:
- Add parseValue to support converting values (eg. string => number). This allows
  better matching when input may come from string source but values are integers.
- Fix handling of useAsyncValue existingValues to work with complex keys

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,14 @@
 
 ## Unreleased
 
+### Added
+
+* Added [AsyncChoices.parseValue](https://prestojs.com/docs/viewmodel/fields/AsyncChoices#Method-parseValue)
+
 ### Changed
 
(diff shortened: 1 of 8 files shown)


fanglong-alliance (approved):
lgtm, but do we want to make parseValue a default thing on it? 

edit: oh that'd be done in template from codegen. can we have it done in template from codegen?

davecoates:
Yeah needs this change in async_chocies.html:

```
@@ -52,4 +52,9 @@ new AsyncChoices<{ {{options.label_field}}: string; {{options.value_field }}: {{
     getValue(item) {
         return item.{{ options.value_field }};
     },
+   {% if options.value_type_info == 'number' %}
+     parseValue(value) {
+        return Number(value);
+     },
+   {% endif %}
 })
```

PR #184, opened 2022-12-07
Fix/boolean field and widget

davecoates:
Review commit by commit
Consolidates BooleanField/NullableBooleanField, makes BooleanWidget more compatible with final-form

Removes a bunch of fields and widgets that provide no extra functionality


js-packages/@prestojs/viewmodel/src/fields/BooleanField.ts
```diff
@@ -22,24 +21,31 @@ import NullableBooleanField from './NullableBooleanField';
  * ```js
  * viewModelFactory({
  *   id: new Field(),
- *   isActive: new CharField({
+ *   isActive: new BooleanField({
```
fanglong-alliance, line 24: 🤣

fanglong-alliance (commented):
lgtm, there's a vercel failure, & i think we want to emphasis in readme that while NullableBooleanField is gone the NullableBooleanWidget is still there

fanglong-alliance (approved):

PR #187, opened 2024-02-28
Migrate useViewModelCache to use useSyncExternalStoreWithSelector

davecoates:
useSyncExternalStoreWithSelector is the official solution for this type of thing (or https://react.dev/reference/react/useSyncExternalStore is, this is a specialised version used in react-redux that seems to be officially supported)

This works better with server side rendering - right now SSR produces many warnings because of the usage of useLayoutEffect (this has no effect on the server). This change gets rid of that and also means data will be available on the server. It worked before anyway because the data is selected outside of useLayoutEffect on initial renders, but this is probably more robust and is more likely to be more compatible with things like concurrent mode going forward.

Also included is adding `deleteAll` to ViewModelCache (tangentially related to issues with SSR in that in AP2 we clear caches in SSR to avoid leaking data between requests)

### DIFF
--- a/js-packages/@prestojs/viewmodel/src/__tests__/useViewModelCache.test.tsx
+++ b/js-packages/@prestojs/viewmodel/src/__tests__/useViewModelCache.test.tsx
@@ -1,5 +1,5 @@
 import deepEqual from 'lodash/isEqual';
-import { act, render, renderHook } from 'presto-testing-library';
+import { act, render, renderHook, renderNoStrictMode } from 'presto-testing-library';
 import React from 'react';
 
 import { recordEqualTo } from '../../../../../js-testing/matchers';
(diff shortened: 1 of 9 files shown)


js-packages/@prestojs/viewmodel/src/__tests__/useViewModelCache.test.tsx
```diff
@@ -66,54 +66,62 @@ test('should handle updates between first render and subscription', () => {
     expect(container.textContent).toBe('Bob, Sam');
 });
 
-test('should rerender with new results on change', () => {
+test('should rerender with new results on change)', () => {
```
micahklaz: Typo: there is an errant ')' at the end of the test description

micahklaz (approved):
One typo but otherwise looks good

